PR #1, opened 2020-03-18
Add extensions.json + two helper scripts for adding and publishing extensions

jankeromnes:
Mostly works, except that publishing always fails with:

```
Error: Unknown publisher: [NAMESPACE]
```

I guess we need to create namespaces when they don't already exist.


jankeromnes:
Merging this without review, but I'm happy to implement any changes as follow-up PRs.

PR #21, opened 2020-04-06
Add helper script to upgrade versions in extensions.json, and use it

jankeromnes:


jankeromnes:
Thanks @akosyakov for the review! 👍 (on Slack)

> it is fine to update
> if old versions are preserved

PR #44, opened 2020-06-17
Add some Ruby and Rails extensions

hari-govind:
Adds the following extensions:
- [Ruby on Rails](https://marketplace.visualstudio.com/items?itemName=hridoy.rails-snippets)
- [Rails](https://marketplace.visualstudio.com/items?itemName=bung87.rails)
- [Simple Ruby ERB](https://marketplace.visualstudio.com/items?itemName=vortizhe.simple-ruby-erb)

### DIFF
--- a/extensions.json
+++ b/extensions.json
@@ -538,6 +538,10 @@
       "repository": "https://github.com/Tom-Bonnike/vscode-formatting-toggle"
     },
     {
+      "id": "vortizhe.simple-ruby-erb",
+      "repository": "https://github.com/vortizhe/vscode-ruby-erb"
+    },
+    {
       "id": "vscode-icons-team.vscode-icons",
       "version": "10.1.1",
       "repository": "https://github.com/vscode-icons/vscode-icons",
--- a/extensions.json
+++ b/extensions.json
@@ -281,6 +281,10 @@
       "repository": "https://github.com/HookyQR/VSCodeBeautify"
     },
     {
+      "id": "hridoy.rails-snippets",
+      "repository": "https://github.com/hridoy/rails-vscode"
+    },
+    {
       "id": "humao.rest-client",
       "version": "0.24.1",
       "repository": "https://github.com/Huachao/vscode-restclient",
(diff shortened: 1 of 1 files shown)


extensions.json
```diff
@@ -276,6 +280,10 @@
       "version": "1.4.11",
       "repository": "https://github.com/HookyQR/VSCodeBeautify"
     },
+    {
+      "id": "hridoy.rails-snippets",
+      "repository": "https://github.com/hridoy/rails-vscode"
```
jankeromnes, line 285: Oops, there is no code or `package.json` file in https://github.com/hridoy/rails-vscode ? How does this work?

hari-govind: Ah, sorry I didn't check that closely. I just used the repository link from the marketplace since it was MIT licensed, I thought the code would be there. Is it good if we could remove that extension for now and proceed with the other two? I'll check this one out with the author and create a new PR for it if possible. I use VSCodium and the recent update only has extension from open-vsx, so trying to get some of the extensions I used over to open-vsx. Thanks :)

jankeromnes: Sure! No problem, I'll merge the other two, and we can see about this one later. 🙂

Thanks for testing the VSCodium/OpenVSX integration, and especially for adding extensions you're missing! That's super helpful 💯

extensions.json
```diff
@@ -276,6 +280,10 @@
...
+    },
```
jankeromnes, line 286:
```suggestion
```

jankeromnes:
Quick updates:

- Publishing `bung87.rails` failed due to `Error: Cannot resolve dependency: bung87.vscode-gemfile` -- so we should also publish the missing dependency `bung87.vscode-gemfile`, but first its license needs to be clarified: https://github.com/bung87/vscode-gemfile/issues/3

- Publishing `vortizhe.simple-ruby-erb` failed due to `Error: Cannot find module '/tmp/repository/node_modules/vscode/bin/install'` but I expect that to be already fixed in https://github.com/open-vsx/publish-extensions/commit/003582f2c815d05234038748c018ec5600043c7a (we'll see tomorrow morning)

PR #45, opened 2020-06-17
Publish vscode-tailwindcss

herrboyer:
Add [bradlc.vscode-tailwindcss](https://marketplace.visualstudio.com/items?itemName=bradlc.vscode-tailwindcss) to `extensions.json`.

Please note that I'm not the maintainer of this extension.

### DIFF
--- a/extensions.json
+++ b/extensions.json
@@ -75,9 +75,10 @@
     },
     {
       "id": "bradlc.vscode-tailwindcss",
-      "repository": "https://github.com/bradlc/vscode-tailwindcss"
+      "version": "0.3.1",
+      "repository": "https://github.com/bradlc/vscode-tailwindcss",
+      "checkout": "v0.3.1"
     },
-
     {
       "id": "bungcip.better-toml",
       "version": "0.3.2",
--- a/extensions.json
+++ b/extensions.json
@@ -74,6 +74,11 @@
       "checkout": "v1.4.1"
     },
     {
+      "id": "bradlc.vscode-tailwindcss",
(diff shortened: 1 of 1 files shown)


extensions.json
```diff
@@ -73,6 +73,11 @@
       "repository": "https://github.com/bmewburn/vscode-intelephense",
       "checkout": "v1.4.1"
     },
+    {
+      "id": "bradlc.vscode-tailwindcss",
+      "repository": "https://github.com/bradlc/vscode-tailwindcss"
+    },
+
```
jankeromnes, line 80: I notice that this repository maintains consistent Git tags (e.g. `v0.3.1`) so I think it's better to publish this extension from the latest released Git tag, as opposed to straight from the `master` branch, using the `checkout` option. (Also, specifying `version` is an optimization that simply skips trying to re-publish the same version every night.)

```suggestion
    {
      "id": "bradlc.vscode-tailwindcss",
      "version": "0.3.1",
      "repository": "https://github.com/bradlc/vscode-tailwindcss",
      "checkout": "v0.3.1"
    },
```

herrboyer: Hi @jankeromnes, thanks for the clarification and the suggestion !

Maybe the README.md could be clearer on this subject, to me it looked like _completely_ optional values, now I understand that we _should_ specify checkout if version tags are available.

jankeromnes:
It worked!

> 🚀  Published bradlc.vscode-tailwindcss v0.4.1

https://open-vsx.org/extension/bradlc/vscode-tailwindcss

PR #46, opened 2020-06-17
docs: recommended values for extensions.json

herrboyer:
This is a follow up to https://github.com/open-vsx/publish-extensions/pull/45#pullrequestreview-432420239 

A quick modification of the `README.md` file to explain why `version` & `checkout` should be provided when adding an extension.

### DIFF
--- a/README.md
+++ b/README.md
@@ -24,7 +24,11 @@ Here is the expected format of an [`extensions.json`](./extensions.json) entry:
       // Unique Open VSX extension ID in the form "<namespace>.<name>"
       "id": "redhat.vscode-yaml",
       // Repository URL to clone and publish from
-      "repository": "https://github.com/redhat-developer/vscode-yaml"
+      "repository": "https://github.com/redhat-developer/vscode-yaml",
+      // (RECOMMENDED) The version to publish to Open VSX (defaults to the package.json version)
+      "version": "0.27.0",
+      // (RECOMMENDED) The Git branch, tag, or commit to check out before publishing (defaults to the repository's default branch)
+      "checkout": "v0.27.0",
     },
 ```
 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -24,7 +24,11 @@ Here is the expected format of an [`extensions.json`](./extensions.json) entry:
       // Unique Open VSX extension ID in the form "<namespace>.<name>"
       "id": "redhat.vscode-yaml",
       // Repository URL to clone and publish from
-      "repository": "https://github.com/redhat-developer/vscode-yaml"
+      "repository": "https://github.com/redhat-developer/vscode-yaml",
+      // (RECOMMENDED) The version to publish to Open VSX (defaults to the package.json version)
+      "version": "0.27.0",
+      // (RECOMMENDED) The Git branch, tag, or commit to check out before publishing (defaults to the repository's default branch)
+      "checkout": "v0.27.0",
```
jankeromnes, line 31: Since repositories don't always have dedicated "release" branches or tags, I prefer keeping `checkout` optional, and also keeping the first example minimal (i.e. only required parameters, since the optional ones are documented below).

```suggestion
      "repository": "https://github.com/redhat-developer/vscode-yaml"
```

README.md
```diff
@@ -34,11 +38,11 @@ Here are all the supported values, including optional ones:
     {
       // Unique Open VSX extension ID in the form "<namespace>.<name>"
       "id": "rebornix.ruby",
-      // (OPTIONAL) The version to publish to Open VSX (defaults to the package.json version)
-      "version": "0.27.0",
       // Repository URL to clone and publish from
       "repository": "https://github.com/rubyide/vscode-ruby",
-      // (OPTIONAL) The Git branch, tag, or commit to check out before publishing (defaults to the repository's default branch)
+      // (RECOMMENDED) The version to publish to Open VSX (defaults to the package.json version)
```
jankeromnes, line 43: Calling it `RECOMMENDED` is a great idea, thanks! 👍 I also like the proposed new options order.

However, my original description was slightly wrong, so I might as well drive-by fix it. 😇

Explanation: This value is only used in comparison with what OpenVSX already has, to know if publishing should be skipped or not (but if publishing happens, it's the actual version string in `package.json` that will be used).

```suggestion
      // (RECOMMENDED) The version that should be published on Open VSX (defaults to the package.json version)
```

README.md
```diff
@@ -34,11 +38,11 @@ Here are all the supported values, including optional ones:
...
+      "version": "0.27.0",
+      // (RECOMMENDED) The Git branch, tag, or commit to check out before publishing (defaults to the repository's default branch), should match "version"
```
jankeromnes, line 45: Due to the reason described above, it's ok to not match "version" (this could also be a "release" or "stale" branch, and you can even omit "version", which just serves to skip publishing when it's not necessary).

```suggestion
      // (RECOMMENDED) The Git branch, tag, or commit to check out before publishing (defaults to the repository's default branch)
```

herrboyer:
Great, thanks again @jankeromnes for adjusting this PR to match your needs !

PR #54, opened 2020-06-21
adding PlatformIO v1.10.0

fbeke:
Add platformio.platformio-ide to extensions.json.
Please note, that I'm not the maintainer of this extension.

### DIFF
--- a/extensions.json
+++ b/extensions.json
@@ -692,7 +692,8 @@
     {
       "id": "platformio.platformio-ide",
       "repository": "https://github.com/platformio/platformio-vscode-ide",
-      "version": "1.10.0"
+      "version": "1.10.0",
+      "checkout": "v1.10.0"
     }
   ]
 }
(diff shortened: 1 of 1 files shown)


extensions.json
```diff
@@ -688,6 +688,11 @@
       "id": "zxh404.vscode-proto3",
       "repository": "https://github.com/zxh0/vscode-proto3",
       "version": "0.4.2"
+    },
+    {
+      "id": "platformio.platformio-ide",
+      "repository": "https://github.com/platformio/platformio-vscode-ide",
+      "version": "1.10.0"
```
jankeromnes, line 695: When specifying a `version`, it's good to also specify a corresponding `checkout` (that way we can be certain that the version we publish is in the state that the maintainer intended it to be, as opposed to publishing it once from `master`)

```suggestion
      "version": "1.10.0",
      "checkout": "v1.10.0"
```

PR #60, opened 2020-06-23
add haml-vscode

demsullivan:


extensions.json
```diff
@@ -373,6 +373,11 @@
       "repository": "https://github.com/styled-components/vscode-styled-components",
       "version": "0.0.29"
     },
+    {
+      "id": "karunamurti.haml",
+      "repository": "https://github.com/karuna/haml-vscode",
+      "version": "1.1.0"
```
jankeromnes, line 379: It seems ok to publish straight from the default branch without pinning a specific version.

Also, I notice that https://github.com/karuna/haml-vscode/blob/25cb1b6ac4e37aebb0b483ea6d5e8264cf5c1eca/package.json#L5 has version `1.1.0` whereas https://marketplace.visualstudio.com/items?itemName=karunamurti.haml has `1.3.0` already... Where is the code for the `1.3.0` version?

```suggestion
      "repository": "https://github.com/karuna/haml-vscode"
```

jankeromnes:
It worked!

> [🚀  Published karunamurti.haml v1.1.0](https://github.com/open-vsx/publish-extensions/runs/820891153?check_suite_focus=true)

https://open-vsx.org/extension/karunamurti/haml

PR #63, opened 2020-06-25
Prepublish github-vscode-theme

simurai:
👋  We got a report that the `github-vscode-theme` doesn't work without running `npm run build` first, see https://github.com/primer/github-vscode-theme/issues/66.

This PR adds the `prepublish` option and hopefully fixes the issue.


jankeromnes:
Merging the fix as it improves the current situation, but moving the OpenVSX publishing to the official repo would be even better.

PR #65, opened 2020-06-28
Add ccls extension

pavel-the-best:


extensions.json
```diff
@@ -105,6 +105,11 @@
       "version": "0.3.2",
       "checkout": "0.3.2"
     },
+    {
+      "id": "ccls-project.ccls",
+      "repository": "https://github.com/MaskRay/vscode-ccls",
+      "version": "0.1.10"
```
jankeromnes, line 111: It seems fine to always publish from the default branch, so no need to pin a specific version (also, the current version seems to be `0.1.28`).

```suggestion
      "repository": "https://github.com/MaskRay/vscode-ccls"
```

jankeromnes:
It worked!

> [🚀  Published ccls-project.ccls v0.1.28](https://github.com/open-vsx/publish-extensions/runs/820891153?check_suite_focus=true)

https://open-vsx.org/extension/ccls-project/ccls

pavel-the-best:
That's great!

PR #76, opened 2020-07-03
add https://github.com/svipas/vscode-control-snippets

lachieh:


extensions.json
```diff
@@ -653,6 +653,11 @@
       "checkout": "v1.9.0",
       "location": "packages/client"
     },
+    {
+      "id": "svipas.control-snippets",
+      "repository": "https://github.com/svipas/vscode-control-snippets",
+      "version": "1.9.0"
```
jankeromnes, line 659: The repo seems to have a dedicated Git tag for this release, so I think it's safer to publish from that instead of from the default branch.

```suggestion
      "version": "1.9.0",
      "checkout": "1.9.0"
```

jankeromnes:
It worked! 🎉 https://open-vsx.org/extension/svipas/control-snippets

lachieh:
Thank you!

PR #85, opened 2020-07-18
add smali extension https://github.com/ViRb3/vscode-smali

54696d21:
not sure if that's the correct way to do it, but I can't find anything obvious that might be wrong: the original repo is under MIT license so I guess it should be fine

### DIFF
--- a/extensions.json
+++ b/extensions.json
@@ -733,8 +733,8 @@
     {
       "id": "vscode.smali",
       "repository": "https://github.com/ViRb3/vscode-smali",
-      "version": "1.0.0",
-      "location": "smali"
+      "location": "smali",
+      "prepublish": "cd smali && npm install"
     },
     {
       "id": "vscodevim.vim",
--- a/extensions.json
+++ b/extensions.json
@@ -731,6 +731,12 @@
       "checkout": "v10.2.0"
     },
     {
+      "id": "vscode.smali",
+      "repository": "https://github.com/ViRb3/vscode-smali",
+      "version": "1.0.0",
+      "location": "smali"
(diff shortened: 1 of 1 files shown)


extensions.json
```diff
@@ -730,6 +730,12 @@
       "version": "10.2.0",
       "checkout": "v10.2.0"
     },
+    {
+      "id": "vscode.smali",
+      "repository": "https://github.com/ViRb3/vscode-smali",
+      "version": "1.0.0",
```
jankeromnes, line 736: Since we're publishing from the default branch (i.e. no `"checkout"` specified), there is no need to pin a specific version. (I.e. if `1.0.1` comes out, it's fine to let the script publish it without manual intervention.)

```suggestion
```

extensions.json
```diff
@@ -730,6 +730,12 @@
...
+      "location": "smali"
```
jankeromnes, line 737: Here the default behavior of of the publishing script may be a bit weird: it always runs `npm install` or `yarn install` in the repo root, and not in the extension location.

To work around this weirdness, I suggest we "also" run `npm install` in the correct location using a `"prepublish"` command:

```suggestion
      "location": "smali",
      "prepublish": "cd smali && npm install"
```

jankeromnes:
Applied suggestions, squashed & merged. Thanks again!

I'll now trigger another publishing run, so that smali gets deployed to https://open-vsx.org -- will try to post back here once it's live.

jankeromnes:
Update: It worked! 🎉 https://open-vsx.org/extension/vscode/smali

Issue #91, opened 2020-07-23
Outdated extensions


acim:
eslint in https://open-vsx.org/vscode/gallery is version 2.1.3 while the latest tag in https://github.com/microsoft/vscode-eslint/tags i 2.1.9.

Same is for the following extensions:

ms-azuretools.vscode-docker 1.0.0 -> 1.4.1
redhat.vscode-yaml 0.8.0 -> 0.9.1

jankeromnes:
Good catch! Many thanks @acim for reporting these problems.

> eslint in https://open-vsx.org/vscode/gallery is version 2.1.3 while the latest tag in https://github.com/microsoft/vscode-eslint/tags i 2.1.9.

The namespace `dbaeumer` was claimed by @dbaeumer in https://github.com/eclipse/open-vsx.org/issues/10.

@dbaeumer could you please publish the latest version of `dbaeumer.vscode-eslint` to https://open-vsx.org?*

---

> ms-azuretools.vscode-docker 1.0.0 -> 1.4.1

The namespace `ms-azuretools` was claimed by @fiveisprime in https://github.com/eclipse/open-vsx.org/issues/5.

@fiveisprime could you please publish the latest version of `ms-azuretools.vscode-docker` to https://open-vsx.org?*

---

> redhat.vscode-yaml 0.8.0 -> 0.9.1

The namespace `redhat` was claimed by @fbricon in https://github.com/eclipse/open-vsx.org/issues/6.

@fbricon could you please publish the latest version of `redhat.vscode-yaml` to https://open-vsx.org?*

---

*Publishing to https://open-vsx.org should be as simple as just running `npx ovsx publish` in your repository (with a `OVSX_PAT` access token that you can get [here](https://open-vsx.org/user-settings/tokens)), but please let me know if you run into issues or would like help publishing your extensions to OpenVSX.

fbricon:
I manually pushed https://open-vsx.org/extension/redhat/vscode-yaml/0.9.1 (cc @JPinkney)

jankeromnes:
Awesome, many thanks @fbricon! 💯

jankeromnes:
Ping @dbaeumer @fiveisprime

jankeromnes:
[...]

PR #94, opened 2020-07-23
Auto-publish ms-azuretools.vscode-docker

mattburgess:


jankeromnes:
And many thanks for this contribution too!

Please note that `ms-azuretools.vscode-docker` is already available on OpenVSX: https://open-vsx.org/extension/ms-azuretools/vscode-docker although in an older version.

While it was previously published from here, the namespace `ms-azuretools` has since been claimed by Microsoft, which means that they will now publish this extension themselves.

We've let them know that `ms-azuretools.vscode-docker` is now outdated, and are awaiting their response in https://github.com/open-vsx/publish-extensions/issues/91#issuecomment-662952787

mattburgess:
Oh dear. I'm not doing very well on my issue searches this evening! Thanks for the lightning quick responses!

jankeromnes:
Many thanks for sending PRs! That's totally awesome in itself. 🙂

PR #95, opened 2020-07-24
Add skn0tt.markdown-link-expander

sepeth:


extensions.json
```diff
@@ -681,6 +681,11 @@
       "version": "1.4.2",
       "checkout": "v1.4.2"
     },
+    {
+      "id": "skn0tt.markdown-link-expander",
+      "repository": "https://github.com/Skn0tt/markdown-link-expander",
+      "version": "0.1.1"
```
jankeromnes, line 687: Since this project seems to have dedicated release tags, let's publish version `0.1.1` from the tag `v0.1.1` instead of from the default branch:

```suggestion
      "version": "0.1.1",
      "checkout": "v0.1.1"
```

PR #102, opened 2020-08-01
Add rokoroku.vscode-theme-darcula

DustInCompetent:


extensions.json
```diff
@@ -644,6 +644,11 @@
       "version": "5.6.1",
       "checkout": "v5.6.1"
     },
+    {
+      "id": "rokoroku.vscode-theme-darcula",
+      "repository": "https://github.com/rokoroku/vscode-theme-darcula",
+      "version": "1.2.1"
```
jankeromnes, line 650: Also, this project has dedicated release tags, so I think it's better to publish version `1.2.1` from the `v1.2.1` tag instead of from the repository's default branch.

```suggestion
      "version": "1.2.1",
      "checkout": "v1.2.1"
```

DustInCompetent:
I've added https://github.com/rokoroku/vscode-theme-darcula/issues/22

DustInCompetent:
@spoenemann They've added an MIT license but didn't tag the commit. Might I set the `"checkout"` setting directly to the latest commit rokoroku/vscode-theme-darcula@4f0a8cf?

jankeromnes:
> They've added an MIT license but didn't tag the commit. Might I set the `"checkout"` value directly to the latest commit rokoroku/vscode-theme-darcula@4f0a8cf?

@DustInCompetent Awesome! And sure, that would be fine, thank you. (Alternatively, you could also remove `"checkout"` and `"version"`, to always publish straight from the default branch, if you think that would be stable enough -- as you prefer.)

jankeromnes:
It worked! 🎉 https://open-vsx.org/extension/rokoroku/vscode-theme-darcula

PR #103, opened 2020-08-03
Updates

jankeromnes:


extensions.json
```diff
@@ -65,7 +65,7 @@
       "version": "1.16.6",
       "checkout": "vscode-apollo@1.16.6",
       "location": "packages/vscode-apollo",
-      "prepublish": "npm --prefix ../.. install && npm run _internal:prepare-extension"
+      "prepublish": "cd packages/vscode-apollo && npm --prefix ../.. install && npm run _internal:prepare-extension"
```
spoenemann, line 68: We could rephrase this as `npm install && npm --prefix packages/vscode-apollo run _internal:prepare-extension`.

jankeromnes: Good idea thanks. But would `npm --prefix packages/vscode-apollo run _internal:prepare-extension` then actually run in the `packages/vscode-apollo` directory, as opposed to just picking up `package.json` from that location but running in the current working directory?

PR #109, opened 2020-08-08
Add Orta.vscode-jest

andreubotella:


jankeromnes:
Hi @andreubotella, many thanks for your contribution! While I'm not against adding `Orta.vscode-jest` to our auto-publish CI here, I think it's generally better when extensions are published to OpenVSX by their original author.

Hi @orta 👋 I notice that you've already published https://open-vsx.org/extension/svelte/svelte-vscode and https://open-vsx.org/extension/svelte/svelte-vscode-nightly to OpenVSX (many thanks for that! 💯)

Would you also be willing to publish https://github.com/jest-community/vscode-jest ?

jankeromnes:
Update: The OpenVSX namespace `Orta` has been claimed in https://github.com/eclipse/open-vsx.org/issues/122, and the extension `Orta.vscode-jest` will get published by its authors in https://github.com/jest-community/vscode-jest/issues/602.

So, I think we can close this for now, as author-publishing is preferred. Thanks again for pushing this forward @andreubotella! 💯

PR #115, opened 2020-08-12
Allow downloading already packaged extensions

spoenemann:
This change adds a `download` option that can be used to download and publish an already packaged extension. This can be useful for cases where it is not feasible to build an extension from source.

### DIFF
--- a/README.md
+++ b/README.md
@@ -36,7 +36,7 @@ Here are all the supported values, including optional ones:
       "id": "rebornix.ruby",
       // Repository URL to clone and publish from
       "repository": "https://github.com/rubyide/vscode-ruby",
-      // (RECOMMENDED) The version that should be published on Open VSX (defaults to the package.json version)
+      // (RECOMMENDED) The version that should be published; the script compares this version with the latest published version
       "version": "0.27.0",
       // (RECOMMENDED) The Git branch, tag, or commit to check out before publishing (defaults to the repository's default branch)
       "checkout": "v0.27.0",
@@ -44,6 +44,19 @@ Here are all the supported values, including optional ones:
       "location": "packages/vscode-ruby-client",
       // (OPTIONAL) Extra commands to run just before publishing to Open VSX (i.e. after "yarn/npm install", but before "vscode:prepublish")
       "prepublish": "npm run build"
+    },
+```
+
+In cases where it is not feasible to build the extension from source, a download URL can be given instead:
+
+```js
+    {
+      // Unique Open VSX extension ID in the form "<namespace>.<name>"
+      "id": "rebornix.ruby",
+      // (RECOMMENDED) The version that should be published; the script compares this version with the latest published version
+      "version": "0.25.0",
+      // A full URL from which to download the extension package
+      "download": "https://github.com/rubyide/vscode-ruby/releases/download/v0.25.0/ruby-0.25.0.vsix",
     },
 ```
 
--- a/README.md
+++ b/README.md
@@ -53,10 +53,10 @@ In cases where it is not feasible to build the extension from source, a download
     {
       // Unique Open VSX extension ID in the form "<namespace>.<name>"
       "id": "rebornix.ruby",
-      // (RECOMMENDED) The version that should be published; the script compares this version with the latest published version
-      "version": "0.25.0",
       // A full URL from which to download the extension package
       "download": "https://github.com/rubyide/vscode-ruby/releases/download/v0.25.0/ruby-0.25.0.vsix",
+      // (RECOMMENDED) The version that should be published; the script compares this version with the latest published version
+      "version": "0.25.0"
     },
 ```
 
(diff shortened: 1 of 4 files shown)


extensions.json
```diff
@@ -585,10 +585,8 @@
     },
     {
       "id": "ms-python.python",
-      "repository": "https://github.com/Microsoft/vscode-python",
       "version": "2020.7.1",
```
akosyakov, line 588: 2020.7.96456 ?

spoenemann: Their tag names are inconsistent with the versions, see https://github.com/microsoft/vscode-python/releases/tag/2020.7.96456

akosyakov: they don't seem to use 2020.7.1 on marketplace: https://marketplace.visualstudio.com/items?itemName=ms-python.python It was confusing to me to understand whether `2020.7.1` corresponds to MS marketplace version or not or to which commit in the repo.

spoenemann: Strange, maybe the last part is replaced in their release process. Switching to prebuilt packages should solve this.

README.md
```diff
@@ -47,6 +47,19 @@ Here are all the supported values, including optional ones:
...
+In cases where it is not feasible to build the extension from source, a download URL can be given instead:
+
+```js
+    {
+      // Unique Open VSX extension ID in the form "<namespace>.<name>"
+      "id": "rebornix.ruby",
+      // (RECOMMENDED) The version that should be published; the script compares this version with the latest published version
+      "version": "0.25.0",
+      // A full URL from which to download the extension package
+      "download": "https://github.com/rubyide/vscode-ruby/releases/download/v0.25.0/ruby-0.25.0.vsix",
```
jankeromnes, line 59: Minor nit: I'd sort the parameters mandatory first / optional last:

```suggestion
      // A full URL from which to download the extension package
      "download": "https://github.com/rubyide/vscode-ruby/releases/download/v0.25.0/ruby-0.25.0.vsix",
      // (RECOMMENDED) The version that should be published; the script compares this version with the latest published version
      "version": "0.25.0"
```

extensions.json
jankeromnes: Keeping `2020.7.1` here will skip publishing (because `2020.7.1` is already published).

I suggest removing the `version` entirely (it's just an optimization to avoid a pointless publishing attempt, removing it guarantees publishing is at least attempted):

```suggestion
```

publish-extensions.js
```diff
@@ -14,11 +14,24 @@ const ovsx = require('ovsx');
...
+   * @type {{
+   *    extensions: {
+   *        id: string,
+   *        repository: string,
+   *        version?: string,
+   *        checkout?: string,
+   *        location?: string,
+   *        prepublish?: string,
+   *        download?: string
+   *    }[]
```
jankeromnes, line 32: Maybe separate the two types to be more precise? (E.g. `repository` is mandatory in one case, but should be omitted in the other).

```suggestion
   *    extensions: {
   *        id: string,
   *        repository: string,
   *        version?: string,
   *        checkout?: string,
   *        location?: string,
   *        prepublish?: string
   *    }[] | {
   *        id: string,
   *        download: string,
   *        version?: string
   *    }[]
```

publish-extensions.js
```diff
@@ -78,19 +75,59 @@ const readFile = util.promisify(fs.readFile);
...
+          console.warn('[WARN] Ignoring `checkout` property because `download` was given.')
+        }
+        if (extension.prepublish) {
+          console.warn('[WARN] Ignoring `prepublish` property because `download` was given.')
+        }
+        if (extension.location) {
+          console.warn('[WARN] Ignoring `location` property because `download` was given.')
+        }
+
+        // Download the extension package from a GitHub release
```
jankeromnes, line 92: The `.vsix` can actually be downloaded from anywhere.

```suggestion
        // Download the extension package, e.g. from a GitHub release
```

publish-extensions.js
```diff
@@ -78,19 +75,59 @@ const readFile = util.promisify(fs.readFile);
...
+        /** @type {import('ovsx').PublishOptions} */
+        const options = { packagePath: path.join('/tmp/repository', extension.location || '.') };
+        if (yarn) {
+            options.yarn = true;
+        }
+        await ovsx.publish(options);
+        console.log(`[OK] Successfully published ${id} to Open VSX!`)
       }
-      await ovsx.publish(options);
-      console.log(`[OK] Successfully published ${id} to Open VSX!`)
```
jankeromnes, line 87: Micro-nit: You could keep this single `console.log` line instead of having it at the bottom of each `if/else` branch.

publish-extensions.js
spoenemann: That's problematic, ts-check then complains when trying to access properties that are only in one of the two alternatives.

PR #121, opened 2020-08-17
adding vscode meson extention

noah1510:


jankeromnes:
Looks like auto-publishing worked! 🎉 https://open-vsx.org/extension/asabil/meson

PR #122, opened 2020-08-18
[justusadam.language-haskell] Add 'make all' for the justusadam.language-haskell extension

bryphe:
__Issue:__ The [language-haskell](https://open-vsx.org/extension/justusadam/language-haskell) plugin is missing syntax files. These are intended to be generated during the build, via `make all`, but this wasn't included in the publish script. This means that the extension is broken - at least for syntax highlighting.

We noticed this in https://github.com/onivim/oni2/issues/2307 using the language-haskell extension published to open-vsx.

__Fix:__ Add `make all` to the prepublish script for `language-haskell`.

Related:
- https://github.com/JustusAdam/language-haskell/issues/171 - tracking publishing by author

### DIFF
--- a/extensions.json
+++ b/extensions.json
@@ -461,7 +461,7 @@
       "repository": "https://github.com/JustusAdam/language-haskell",
       "version": "3.3.0",
       "checkout": "v3.3.0",
-      "prepublish": "npm i --save-dev @types/node@10"
+      "prepublish": "npm i --save-dev @types/node@10 && make all"
     },
     {
       "id": "karunamurti.haml",


extensions.json
```diff
@@ -461,7 +461,7 @@
       "repository": "https://github.com/JustusAdam/language-haskell",
       "version": "3.3.0",
       "checkout": "v3.3.0",
-      "prepublish": "npm i --save-dev @types/node@10"
+      "prepublish": "npm i --save-dev @types/node@10 && make all"
```
bryphe, line 464: I tested this locally and worked - and it seems like the `ubuntu-latest` image contains `build-essentials` package which has `make`. But I wasn't able to test it in the GH workflow environment. 

Would be cool if there was a 'dry run' for PR validation! 😄

bryphe:
Also some notes about the generation here: https://github.com/JustusAdam/language-haskell/issues/169

jankeromnes (approved):
Wow, great catch, many thanks for fixing this! 💯 

Fix looks good to me, so I'll merge it and report back CI results. (I agree that a PR validation mechanism would be great -- albeit one that doesn't actually upload stuff to OpenVSX!)

I've also subscribed to the issue about publishing this extension upstream. Thanks again!

spoenemann:
I cannot remove the current version 3.3.0 because it is the only published version, and `haskell.haskell` depends on this extension. We need to publish at least one other version before I can remove the broken one.

jankeromnes:
@spoenemann Maybe we can circumvent this, e.g. by publishing an older version, that way you can delete the broken upload, and we re-publish the latest again.

jankeromnes:
Oops, publishing tag `v3.0.3` failed with:

```
Running: npm i --save-dev @types/node@10 && make all
  "id": "justusadam.language-haskell",
  "repository": "https://github.com/JustusAdam/language-haskell",
  "checkout": "v3.0.3",
  "prepublish": "npm i --save-dev @types/node@10 && make all"
}
Error: Command failed: npm i --save-dev @types/node@10 && make all
make: *** No rule to make target 'all'.  Stop.
```

I guess this `all` rule is quite new?

bryphe:
Thanks @jankeromnes @spoenemann !

> Oops, publishing tag v3.0.3 failed

Ah, ya, it looks like the `make` step was added after the `v3.0.3` tag - so you'd need to publish 3.0.3 without the `make all` step

jankeromnes:
> 🚀  Published justusadam.language-haskell v3.0.3

@spoenemann Please delete version 3.3.0 of https://open-vsx.org/extension/justusadam/language-haskell so that we can re-publish a fixed version. 🙂

imsekun:
The issue still persists

Error

```
(Sys_error
  "/home/user/.config/oni2/extensions/justusadam.language-haskell-3.0.3.vsix/syntaxes/haskell.json: No such file or directory"):
```

Steps:

1. Install extension

    - Through extensions tab `v3.3.0`
    - Through open-vsx (with uninstalling previous) `v3.0.3`

2. Open `.hs` file
3. Crashes

Issue #37, opened 2020-06-14
[alexcvzz.vscode-sqlite] sh: 1: preact: not found


jankeromnes:
```
> vscode-sqlite.htmlcontent@0.0.0 build /tmp/repository/src/resultview/htmlcontent
> preact build --dest ../../../out/resultview/htmlcontent

sh: 1: preact: not found
npm ERR! code ELIFECYCLE
npm ERR! syscall spawn
npm ERR! file sh
npm ERR! errno ENOENT
npm ERR! vscode-sqlite.htmlcontent@0.0.0 build: `preact build --dest ../../../out/resultview/htmlcontent`
npm ERR! spawn ENOENT
```

```
[FAIL] Could not process extension: {
  "id": "alexcvzz.vscode-sqlite",
  "version": "0.8.2",
  "repository": "https://github.com/AlexCovizzi/vscode-sqlite"
}
```

jankeromnes:
Not actually fixed.

jankeromnes:
So, `npm install preact` also didn't work. Publishing still fails with:

```
sh: 1: preact: not found
```

jankeromnes:
Ah, rats, the package is actually called `preact-cli`, not `preact`. 🤦

PR #125, opened 2020-08-19
Add SQLite

moodduckk:
It is under Apache-2.0 License

### DIFF
--- a/extensions.json
+++ b/extensions.json
@@ -934,6 +934,11 @@
       "id": "zxh404.vscode-proto3",
       "repository": "https://github.com/zxh0/vscode-proto3",
       "version": "0.4.2"
+    },
+    {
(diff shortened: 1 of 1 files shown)


jankeromnes (commented):
Hi @moodduckk, many thanks for suggesting this extension!

However, it is already present in `extensions.json`:

https://github.com/open-vsx/publish-extensions/blob/83ce6390fcc2f39feb5927032ca9fa3c4e8cb136/extensions.json#L34-L38

(If you try to add it with the script `node add-extension https://github.com/AlexCovizzi/vscode-sqlite, it should tell you that.)

However, the extension is not yet present on https://open-vsx.org, because publishing currently fails: https://github.com/open-vsx/publish-extensions/issues/37

Issue #49, opened 2020-06-18
Add hadolint (GPL-3.0)


Kreyren:
https://github.com/hadolint/hadolint

https://marketplace.visualstudio.com/items?itemName=exiasr.hadolint

Kreyren:
cc @jankeromnes

jankeromnes:
Interesting, thanks!

But I don't think that GPL-3 licensed extensions can be used on Open VSX yet -- @spoenemann do you know?

spoenemann:
Right. We haven't built in any technical restriction regarding the license yet, but we'll do that as the transfer to the Eclipse Foundation gets forward. The admissible licenses are Apache-2.0, EPL-2.0, MIT, ISC, BSD-2-Clause, BSD-3-Clause. But you can contact license@eclipse.org to determine if another license, like GPL-3, is acceptable.

Kreyren:
Why wouldn't the license that basically built the open-source be acceptable here? :thinking:

spoenemann:
I don't know, but I'm sure the Eclipse Foundation has good reasons for it. You could ask via the email address posted above.

Kreyren:
Email sent 

On June 19, 2020 11:10:49 AM GMT+02:00, "Miro Spönemann" <notifications@github.com> wrote:
>I don't know, but I'm sure the Eclipse Foundation has good reasons for
>it. You could ask via the email address posted above.
>
>-- 
>You are receiving this because you authored the thread.
>Reply to this email directly or view it on GitHub:
>https://github.com/open-vsx/publish-extensions/issues/49#issuecomment-646527007

-- 
Sent from my Android device with K-9 Mail. Please excuse my brevity.

jankeromnes:
Since we haven't heard back yet, I've also asked about GPL-3.0 (since I was emailing Eclipse about another license anyway).

This is the official response:

> Thanks for checking in with us.  Unfortunately no, the GPL (any version) is not acceptable because of license incompatibility reasons. Many years ago the FSF and the Eclipse Foundation jointly [published guidance](https://eclipse-foundation.blog/2010/04/06/epl-gpl-commentary/) regarding the use of the GPL for plugins, and this guidance would apply to OpenVSX extensions as well.  
> 
> The following are the Eclipse Board Approved license information for OpenVSX extensions:
> 
> Registry Code Content Offerings will be subject to one or more of the following licenses:
> 
[...]

PR #131, opened 2020-08-22
Add Zenburn, and 2 OCaml extensions

edwintorok:


spoenemann:
Thanks for the contribution!

`hc-zenburn-vscode` has no license. I opened https://github.com/drzix/hc-zenburn-vscode/issues/4 to address this, but we have to wait until that is done before we can add it to Open VSX.

`vscode-reasonml` says in its README that it's currently unmaintained. Do you think it is still useful?

Generally it would be good to add `checkout` properties with the respective release tags to all new entries. This ensures that we build from exactly the sources that were used for the version we want to install.

edwintorok:
Indeed vscode-reasonml is deprecated and doesn't work with recent versions of OCaml anymore, I removed it from this PR.
There is an alternative linked from its github page, and I opened an issue to get it added to open-vsx.org: https://github.com/jaredly/reason-language-server/issues/465. It is not as simple as adding to package.json because package.json is missing some fields.

edwintorok:
A license file has been added for hc-zenburn-vscode, and I've added a checkout property for ocamllabs.ocaml-platform,  so hopefully this PR is good to go now? (hc-zenburn-vscode has no tags, so I could only add a hash as checkout, not sure if that'd be useful)

spoenemann:
@drzix chose GPL-3.0 for hc-zenburn-vscode. Unfortunately that license is not compatible with requirements of the Eclipse Foundation at this moment, see https://github.com/open-vsx/publish-extensions/issues/49#issuecomment-657778606.

@edwintorok could you remove that extension from this PR? We might be able to include GPL licensed extensions at some point, for now we should stick with the current requirements.

edwintorok:
I intended to use hc-zenburn-vscode with Codium which is MIT licensed and therefore compatible with GPL3. It'd be nice if open-vsix had a way of accepting GPL3 extensions for the purposes of Codium (and hide it from Eclipse if it has to).

Anyway I've replaced that plugin with another one that is MIT licensed (it is not the high-contrast version).

spoenemann (approved):

spoenemann:
> It'd be nice if open-vsix had a way of accepting GPL3 extensions for the purposes of Codium

I agree. We might get back to this in case the requirements will change.

jankeromnes:
Update: Auto-publishing worked! 🎉 
- https://open-vsx.org/extension/ocamllabs/ocaml-platform
- https://open-vsx.org/extension/ryanolsonx/zenburn

> It'd be nice if open-vsix had a way of accepting GPL3 extensions for the purposes of Codium

I agree as well. FYI, we're currently investigating a solution that could potentially allow any open-source license onto OpenVSX. If it works, you can expect more news on that in a few weeks or so.

PR #145, opened 2020-09-03
Add Argdown extension

im-0:
See https://argdown.org/ for details.

I'm not sure that hack with `rm -f .yarnrc` in "checkout" will actually work. Is there any way to test this before merging?

### DIFF
--- a/extensions.json
+++ b/extensions.json
@@ -164,6 +164,13 @@
       "version": "2.2.1"
     },
     {
+      "id": "christianvoigt.argdown-vscode",
+      "repository": "https://github.com/christianvoigt/argdown",
+      "version": "1.5.0",
+      "checkout": "be86c880e39da3d963b2f63855f30616a1704703",
+      "location": "packages/argdown-vscode"
+    },
+    {
       "id": "CoenraadS.bracket-pair-colorizer",
       "repository": "https://github.com/CoenraadS/BracketPair",
       "version": "1.0.61"


extensions.json
```diff
@@ -163,6 +163,13 @@
       "repository": "https://github.com/ChristianKohler/PathIntellisense",
       "version": "2.2.1"
     },
+    {
+      "id": "christianvoigt.argdown-vscode",
+      "repository": "https://github.com/christianvoigt/argdown",
+      "version": "1.5.0",
+      "checkout": "argdown-vscode@1.5.0 && rm -f .yarnrc",
```
jankeromnes, line 170: > I'm not sure that hack with `rm -f .yarnrc` in "checkout" will actually work.

Ha, nice one, it seems like your trick would actually work:

https://github.com/open-vsx/publish-extensions/blob/0be92c462d77e884aae5f49af127831f5d471ae9/publish-extensions.js#L104-L106

The preferred, more "official" way to do these things is with a dedicated `"prepublish"` option:

https://github.com/open-vsx/publish-extensions/blob/0be92c462d77e884aae5f49af127831f5d471ae9/publish-extensions.js#L111-L113

However, what is the purpose of deleting `.yarnrc`? I guess that, if you want to delete it before the `yarn install` happens, the `"prepublish"` option would indeed come in too late.

> Is there any way to test this before merging?

Unfortunately, not really. You could approximate a publish run locally by keeping only your extension in `extensions.json` and running `node publish-extensions`, but this isn't ideal because it will fail on OpenVSX calls (needs a `OVSX_PAT` token) and will actually attempt to publish for real.

But I'm fine with merging untested PRs, and then dealing with any resulting CI failures myself (and if that becomes problematic, I'll be encouraged to implement a better way to test PRs).

jankeromnes (commented):
Other than the `checkout` hack, your PR looks good to me. 👍 

I'm just unsure about whether we should implement a `preinstall` option just for this extension (sounds like unnecessary extra complexity in our code), or whether we should just merge the hack and live with it (could come back to bite us in the future, e.g. if we need `checkout` to be a pure Git ref / we start using it elsewhere).

extensions.json
im-0: > However, what is the purpose of deleting `.yarnrc`?

`.yarnrc` from the repository contains this:

```
yarn-path ".yarn/releases/yarn-1.22.4.js"
```

which breaks everything as yarn itself is not included in the repository.

> I guess that, if you want to delete it before the `yarn install happens`, the `"prepublish"` option would indeed come in too late.

Right.

im-0: Thought just a moment ago: maybe that file was pushed by mistake in the first place? I'll try to send pull request to the upstream.

jankeromnes:
I guess cleaner is better. @im-0 would you be willing to implement a `extension.preinstall` option in `publish-extensions.js`, to be handled between `extension.checkout` and the Yarn detection here:

https://github.com/open-vsx/publish-extensions/blob/668f68ce4beef648c8b0ed665761ad25d7f33254/publish-extensions.js#L104-L109

But if you don't have time, I'm happy to merge your PR as is and follow-up with a refactor.

im-0:
Let's wait for answer here: https://github.com/christianvoigt/argdown/pull/188

im-0:
Author of argdown fixed the problem with `.yarnrc`. Special hack is not needed anymore.

jankeromnes (approved):
Awesome! Looks good to me then. Many thanks! 👍

PR #146, opened 2020-09-03
Fix vetur extension: compiled javascript was not included in bundle

bryphe:
The [`vetur`](https://github.com/vuejs/vetur) extension has build steps that aren't included in the current `extensions.json`, so it was missing some build output necessary for running.

Came across as this error during runtime:
image

This uses the download-vsix strategy instead, which includes the build output needed for activation

### DIFF
--- a/extensions.json
+++ b/extensions.json
@@ -675,9 +675,8 @@
     },
     {
       "id": "octref.vetur",
-      "repository": "https://github.com/vuejs/vetur",
-      "version": "0.26.1",
-      "checkout": "v0.26.1"
+      "download": "https://github.com/vuejs/vetur/releases/download/v0.27.0/vetur-0.27.0.vsix",
+      "version": "0.27.0"
     },
     {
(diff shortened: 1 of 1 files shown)


jankeromnes (approved):
Awesome! Many thanks for catching that.

And you're right, re-publishing official `.vsix` files is preferred when available, because trying to guess the correct packaging steps for all extensions in this repo has proven to be error-prone.

jankeromnes:
The fixed version `0.27.0` is now live: https://open-vsx.org/extension/octref/vetur 🎉 

@bryphe Should we delete the older versions? If they don't work, we might as well delete them from OpenVSX.

bryphe:
Thanks @jankeromnes !

> @bryphe Should we delete the older versions? If they don't work, we might as well delete them from OpenVSX.

Deleting the older, broken versions make sense to me - will keep users from accidentally picking up a version that doesn't work.

spoenemann:
Deleted.

PR #147, opened 2020-09-04
Add lextudio.restructuredtext to extensions.json

tchernobog:


extensions.json
```diff
@@ -518,6 +518,10 @@
       "version": "0.6.6",
       "checkout": "v0.6.6"
     },
+    {
+      "id": "lextudio.restructuredtext",
+      "repository": "https://github.com/vscode-restructuredtext/vscode-restructuredtext"
```
jankeromnes, line 523: This project seems to maintain dedicated release tags, so I suggest we use them instead of publishing straight from the default branch.

```suggestion
      "repository": "https://github.com/vscode-restructuredtext/vscode-restructuredtext",
      "version": "129.0.0",
      "checkout": "129.0.0"
```

jankeromnes (approved):
Looks good to me! Many thanks for your contribution.

Two minor things:

- The `displayName` and `description` fiels in the `package.json` seem to be templated somehow: https://github.com/vscode-restructuredtext/vscode-restructuredtext/blob/129.0.0/package.json#L3-L4 (I hope our auto-publishing system will do that correctly, or we risk publishing this extension with a bad description / displayName)

- I suggest we use the dedicated release tags (see in-line comment)

jankeromnes:
Auto-publishing worked! 🎉 https://open-vsx.org/extension/lextudio/restructuredtext

PR #150, opened 2020-09-05
Fix Argdown

im-0:
Currently it fails with

```
sh: 1: rimraf: not found
```

(https://github.com/open-vsx/publish-extensions/runs/1074685213)


spoenemann (approved):
Thanks!

jankeromnes:
You've probably noticed already, but the fix worked! 🎉 https://open-vsx.org/extension/christianvoigt/argdown-vscode

Thanks again. 🙏

PR #155, opened 2020-09-09
Update prettier version.

alexventuraio:


jankeromnes (changes requested):
Cool, many thanks for catching this!

However, this won't work because the vsix download URL still points to the old version.

Side note: you can open this repository in Gitpod, and run `node upgrade-extensions` to update extension versions semi-automatically (and then use `git add -p extensions` to only commit the changes that seem correct/desirable).

alexventuraio:
Jajaja 😅 sorry for that @jankeromnes, I just wanted to move fast jajaja!
I've just fixed it!

And, I have never used `Gitpod` so I will do some research to use it, thanks! 🙏

jankeromnes (approved):
Looks good to me, many thanks!

jankeromnes:
@alexventuraio You've probably noticed already, but the update worked! 🎉 https://open-vsx.org/extension/esbenp/prettier-vscode

Thanks again 👍

alexventuraio:
Wohoo! Great! 👏🏻

PR #156, opened 2020-09-11
Add nightswitch-lite and auto-dark-mode light/dark modes toggle

cd4d:
Extensions that automatically toggle light/dark themes: 

[Nightswitch-lite](https://marketplace.visualstudio.com/items?itemName=gharveymn.nightswitch-lite) based on geolocation

[Auto dark mode](https://marketplace.visualstudio.com/items?itemName=danielgjackson.auto-dark-mode-windows&ssr=false) based on OS theme

Both have MIT licenses.

### DIFF
--- a/extensions.json
+++ b/extensions.json
@@ -211,9 +211,7 @@
     },
     {
       "id": "danielgjackson.auto-dark-mode-windows",
-      "repository": "https://github.com/danielgjackson/vscode-auto-dark-mode-windows",
-      "version": "1.0.8",
-      "checkout": "v1.0.8"
+      "repository": "https://github.com/danielgjackson/vscode-auto-dark-mode-windows"
     },
     
     {
@@ -396,9 +394,7 @@
     },
     {
       "id": "gharveymn.nightswitch-lite",
-      "repository": "https://github.com/gharveymn/nightswitch-lite",
-      "version": "2.2.2",
-      "checkout": "v2.2.2"
+      "repository": "https://github.com/gharveymn/nightswitch-lite"
     },
     {
       "id": "giscafer.leek-fund",
--- a/extensions.json
+++ b/extensions.json
@@ -210,6 +210,13 @@
       "repository": "https://github.com/csstools/postcss-language"
     },
     {
+      "id": "danielgjackson.auto-dark-mode-windows",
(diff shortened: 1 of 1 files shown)


extensions.json
```diff
@@ -209,6 +209,13 @@
       "id": "csstools.postcss",
       "repository": "https://github.com/csstools/postcss-language"
     },
+    {
+      "id": "danielgjackson.auto-dark-mode-windows",
+      "repository": "https://github.com/danielgjackson/vscode-auto-dark-mode-windows",
+      "version": "1.0.8",
+      "checkout": "v1.0.8"
```
jankeromnes, line 216: Unfortunately, there doesn't seem to be a `v1.0.8` tag for this repository: https://github.com/danielgjackson/vscode-auto-dark-mode-windows/tree/v1.0.8

Maybe there is no need to pin a specific version for this repository:

```suggestion
      "repository": "https://github.com/danielgjackson/vscode-auto-dark-mode-windows"
```

extensions.json
```diff
@@ -387,6 +394,12 @@
       "id": "neikeq.godot-csharp-vscode",
       "repository": "https://github.com/godotengine/godot-csharp-vscode"
     },
+    {
+      "id": "gharveymn.nightswitch-lite",
+      "repository": "https://github.com/gharveymn/nightswitch-lite",
+      "version": "2.2.2",
+      "checkout": "v2.2.2"
```
jankeromnes, line 401: Same here, there is no `v2.2.2` tag in this repository.

Even weirder, the repository is still at version `2.2.1`: https://github.com/gharveymn/nightswitch-lite/blob/master/package.json even though the VSCode marketplace already has `2.2.2` -- where is the code for version `2.2.2`?

In any case, maybe let's not pin a specific version for this extension either, i.e. let's always auto-publish from the default branch as that's the best we can do:

```suggestion
      "repository": "https://github.com/gharveymn/nightswitch-lite"
```

jankeromnes (approved):
Many thanks for this contribution!

Looks good to me, but without the `"checkout"`, because the specified Git tags don't actually exist.

Note: I can apply my own suggestions directly in GitHub, so I'll do that an then merge your PR without further bothering you. 🙂

cd4d:
Thanks!

PR #160, opened 2020-09-17
add ms-vscode.js-debug and support extensionFile property

amiramw:


GitMensch:
Yay, that one actually doesn't have the bad MS license but [a plain MIT](https://marketplace.visualstudio.com/items/ms-vscode.js-debug/license). @amiramw Did you already checked if it works without "the Visual Studio Code product", for example with vscodium or even with theia when installed from vsix?

amiramw:
I tested it on theia

spoenemann (commented):
Thanks for the contribution! Three remarks:
 * The `extensionFile` option cannot be used together with `packagePath` (see `ovsx` [cli code](https://github.com/eclipse/openvsx/blob/7074783f8d093036e6369d4f050f3ec97b161b6f/cli/src/publish.ts#L56-L64)), so the latter must be omitted when the first is given.
 * Please add `extensionFile` to the type definition in lines 23-33.
 * Please document the new property in README.md (mark it as _ OPTIONAL_, see `location` and `prepublish`).

spoenemann (approved):
Great, thanks!

jankeromnes:
Update: Auto-publishing worked! 🎉 https://open-vsx.org/extension/ms-vscode/js-debug

PR #164, opened 2020-09-24
added VHDL by VHDLwhiz 

averageunicorn:


spoenemann (commented):
Thanks for the contribution!

When you specify a `download` URL, you have to provide a downloadable .vsix file, for example attached to a tag / release on GitHub. The `awesome-vhdl` project doesn't have any .vsix files in its tags, so the only way to publish it is to build it from source. Please use the `repository` property, and possibly more properties as described in [README.md](https://github.com/open-vsx/publish-extensions#publishing-options).

averageunicorn:
Ah, thank you for the feedback, I'll try to add those.

I'm not sure I understand this correctly, though, possibly because I don't fully understand the process here. Does that mean that `awesome-vhdl` can't be added at all, or only that additional properties are needed?

averageunicorn:
Ah, sorry – I think I fixed these things now, but apparently I set a rubbish commit email address a while ago that got this linked to a different account. This is the first time I'm really-actually-seriously trying to use git and I'm learning a lot, thanks for bearing with me.

extensions.json
```diff
@@ -792,6 +792,18 @@
       "version": "3.0.6",
       "checkout": "v3.0.6"
     },
+    {
+      "id": "ms-vscode.cpptools",
+      "repository": "https://github.com/Microsoft/vscode-cpptools",
```
spoenemann, line 797: The `cpptools` extension has a license that makes it incompatible with anything except Microsoft products:

https://github.com/microsoft/vscode-cpptools/blob/master/RuntimeLicenses/cpptools-LICENSE.txt

Please remove that extension.

extensions.json
```diff
@@ -1047,6 +1059,11 @@
       "id": "twxs.cmake",
       "repository": "https://github.com/twxs/vs.language.cmake"
     },
+    {
+      "id": "vhdlwhiz.vhdl-by-vhdlwhiz",
+      "repository": "https://github.com/jonasjj/awesome-vhdl",
+      "version": "1.2.4"
```
spoenemann, line 1065:
```suggestion
      "version": "1.2.4",
      "checkout": "v1.2.4"
```

This is to ensure that we build exactly from the sources of the 1.2.4 version.

spoenemann:
@averageunicorn could you also add the `checkout` property as shown in the comment above?

averageunicorn:
> @averageunicorn could you also add the `checkout` property as shown in the comment above?

Thank you for reminding me. I think I did now.

spoenemann (approved):
Thanks!

jankeromnes:
FYI, publishing was successful! 🎉 https://open-vsx.org/extension/vhdlwhiz/vhdl-by-vhdlwhiz

PR #167, opened 2020-09-25
Add deibit.devdocs

tywmick:
[devdocs](https://marketplace.visualstudio.com/items?itemName=deibit.devdocs) is an MIT-licensed extension that adds a command to quickly search for selected text in [DevDocs](https://devdocs.io/), automatically inferring the language/framework search context from the file type.

While the repo doesn't have any git tags and its [`package.json`](https://github.com/deibit/vscode-devdocs/blob/master/package.json) doesn't include any packaging or publishing scripts, I tried cloning it myself and running

```zsh
vsce package
code --install-extension devdocs-0.2.0.vsix
```

...and it installed a perfectly functioning extension in my VSCodium installation, so we should have everything we need. 👍🏼

### DIFF
--- a/extensions.json
+++ b/extensions.json
@@ -244,6 +244,11 @@
       "version": "2.1.8"
     },
     {
+      "id": "deibit.devdocs",
+      "repository": "https://github.com/deibit/vscode-devdocs",
+      "version": "0.2.0"
+    },
+    {
       "id": "denoland.vscode-deno",
       "repository": "https://github.com/denoland/vscode_deno"
     },
--- a/extensions.json
+++ b/extensions.json
@@ -245,8 +245,7 @@
     },
     {
       "id": "deibit.devdocs",
-      "repository": "https://github.com/deibit/vscode-devdocs",
-      "version": "0.2.0"
+      "repository": "https://github.com/deibit/vscode-devdocs"
     },
     {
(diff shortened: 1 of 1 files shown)


extensions.json
```diff
@@ -243,6 +243,11 @@
       "download": "https://github.com/microsoft/vscode-eslint/releases/download/release%2F2.1.8/dbaeumer.vscode-eslint-2.1.8.vsix",
       "version": "2.1.8"
     },
+    {
+      "id": "deibit.devdocs",
+      "repository": "https://github.com/deibit/vscode-devdocs",
+      "version": "0.2.0"
```
jankeromnes, line 249: > the repo doesn't have any git tags and its [`package.json`](https://github.com/deibit/vscode-devdocs/blob/master/package.json) doesn't include any packaging or publishing scripts

In that case, we don't need to pin a specific version, as we can always publish the latest version from the default branch.

```suggestion
      "repository": "https://github.com/deibit/vscode-devdocs"
```

jankeromnes (approved):
Looks good to me! Many thanks for the contribution. 👍

jankeromnes:
Update: Auto-publishing worked! 🎉 https://open-vsx.org/extension/deibit/devdocs

PR #168, opened 2020-09-27
Bump extensions hosted on GitHub

niklaskeerl:
This commit bumps all extensions hosted on GitHub to the newest version.

### DIFF
--- a/extensions.json
+++ b/extensions.json
@@ -91,8 +91,8 @@
     {
       "id": "asvetliakov.vscode-neovim",
       "repository": "https://github.com/asvetliakov/vscode-neovim",
-      "version": "0.0.52",
-      "checkout": "v0.0.52"
+      "version": "0.0.63",
+      "checkout": "v0.0.63"
     },
     {
       "id": "atishay-jain.all-autocomplete",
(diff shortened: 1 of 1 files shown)


extensions.json
```diff
@@ -900,8 +900,8 @@
     {
       "id": "platformio.platformio-ide",
       "repository": "https://github.com/platformio/platformio-vscode-ide",
```
jankeromnes, line 902: I notice that this extension made a pre-packaged `.vsix` version available.

In these cases, we prefer re-publishing official `.vsix` files rather than trying to re-package it ourselves (bears the risk of doing something wrong and uploading a broken extension to OpenVSX, e.g. if we forget an important packaging step).

```suggestion
      "download": "https://github.com/platformio/platformio-vscode-ide/releases/download/v2.1.0/platformio-ide-2.1.0.vsix",
```

extensions.json
```diff
@@ -900,8 +900,8 @@
...
-      "version": "1.10.0",
-      "checkout": "v1.10.0"
+      "version": "2.1.0",
+      "checkout": "v2.1.0"
```
jankeromnes, line 904: When using `"download"` instead of `"repository"` (see suggestion above), we no longer need to use `"checkout"`:

```suggestion
      "version": "2.1.0"
```

jankeromnes (approved):
Wow, that's awesome, many thanks @niklaskeerl! 💯 That's super helpful.

Out of curiosity, did you use the `node upgrade-extensions` script?

niklaskeerl:
> Out of curiosity, did you use the `node upgrade-extensions` script?

No, I wrote a script to check if each extension has a new tag on GitHub.

```
let Parser = require("rss-parser");
let parser = new Parser();
var extensions = require("./extensions.json");

extensions.extensions.forEach((item) => {
  if (
    item.repository == null ||
    item.version == null ||
    item.checkout == null
  ) {
  } else {
    repo_url = item.repository.replace(".git", "");
    parser
      .parseURL(repo_url + "/tags.atom")
      .then((feed) => {
        if (item.checkout != feed.items[0].link.split("/").pop()) {
          console.log(item.id);
          console.log(item.checkout);
          console.log(feed.items[0].link.split("/").pop());
        }
      })
      .catch((err) => {
        console.log(item.repository + "/tags.atom");
        console.log(err);
      });
  }
});
```

PR #169, opened 2020-09-27
Update GitHub.vscode-pull-request-github to 0.20.0

Glandos:


extensions.json
```diff
@@ -421,8 +421,8 @@
     {
       "id": "GitHub.vscode-pull-request-github",
       "repository": "https://github.com/microsoft/vscode-pull-request-github",
```
jankeromnes, line 423: I notice that this extension made a pre-packaged `.vsix` version available.

In these cases, we prefer re-publishing official `.vsix` files rather than trying to re-package it ourselves (bears the risk of doing something wrong and uploading a broken extension to OpenVSX, e.g. if we forget an important packaging step).

```suggestion
      "download": "https://github.com/microsoft/vscode-pull-request-github/releases/download/0.20.0/vscode-pull-request-github-0.20.0.vsix",
```

extensions.json
```diff
@@ -421,8 +421,8 @@
...
-      "version": "0.19.0",
-      "checkout": "0.19.0"
+      "version": "0.20.0",
+      "checkout": "0.20.0"
```
jankeromnes, line 425: When using `"download"` instead of `"repository"` (see suggestion above), we no longer need to use `"checkout"`:

```suggestion
      "version": "0.20.0"
```

jankeromnes (approved):
Awesome, many thanks for catching this!

I've made one suggestion in-line (about preferring official `.vsix` files over re-packaging ourselves), but I'll be able to apply it myself before merging your Pull Request. 👍

PR #173, opened 2020-10-01
Added cmake-tools and cpp-tools to extensions

kitattyor:
**cmake-tools** : Easily configure and build cmake based projects.

**cpp-tools**: intellisense and debugger frontend. Is platform specific, hence the multiple versions. Integrates well with cmake-tools.

### DIFF
--- a/extensions.json
+++ b/extensions.json
@@ -659,6 +659,10 @@
       "repository": "https://github.com/BeastCode/VSCode-Angular-TypeScript-Snippets"
     },
     {
+      "id": "microsoft.vscode-cmake-tools",
+      "repository": "https://github.com/microsoft/vscode-cmake-tools"
+    },
+    {
       "id": "mikestead.dotenv",
       "repository": "https://github.com/mikestead/vscode-dotenv",
       "version": "1.0.1"
--- a/extensions.json
+++ b/extensions.json
@@ -663,6 +663,31 @@
       "repository": "https://github.com/microsoft/vscode-cmake-tools"
     },
     {
+      "id": "microsoft.vscode-cpptools-linux",
+      "download": "https://github.com/microsoft/vscode-cpptools/releases/download/1.0.1/cpptools-linux.vsix",
+      "version": "1.0.1"
+    },
+    {
+      "id": "microsoft.vscode-cpptools-linuxARM32",
+      "download": "https://github.com/microsoft/vscode-cpptools/releases/download/1.0.1/cpptools-linux-armhf.vsix",
+      "version": "1.0.1"
+    },
+    {
+      "id": "microsoft.vscode-cpptools-linuxARM64",
+      "download": "https://github.com/microsoft/vscode-cpptools/releases/download/1.0.1/cpptools-linux-aarch64.vsix",
+      "version": "1.0.1"
+    },
+    {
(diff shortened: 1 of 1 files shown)


spoenemann:
The license of cpptools makes it impossible to use the extension in anything except Microsoft products, so we cannot accept the extension here:
https://github.com/microsoft/vscode-cpptools/blob/master/RuntimeLicenses/cpptools-LICENSE.txt

cmake-tools looks ok license-wise, but it has a dependency on cpptools:
https://github.com/microsoft/vscode-cmake-tools/blob/80a745aeb7a12c76f5976a7b4ebca9674d17b5dd/package.json#L1623

herandol:
Does the RuntimeLicenses only apply to the .vsix provided by Microsoft or to all builds ?
I was thinking to build the extension in prepublish.

spoenemann:
The extension also has runtime dependencies that it downloads on startup:

https://github.com/microsoft/vscode-cpptools/blob/9650bb7d9fba86d454847447aecb2450585ff330/Extension/package.json#L2290-L2527

I haven't checked, but I guess those are proprietary, too.

spoenemann:
> Does the RuntimeLicenses only apply to the .vsix provided by Microsoft or to all builds ?

I don't know, you should ask the maintainers.

kitattyor:
> 
> 
> The extension also has runtime dependencies that it downloads on startup:
> 
> https://github.com/microsoft/vscode-cpptools/blob/9650bb7d9fba86d454847447aecb2450585ff330/Extension/package.json#L2290-L2527
> 
> I haven't checked, but I guess those are proprietary, too.

Yes I now see that vscode-cpptools seems to be a MIT licensed wrapper around the cpp language service which cannot be used on anything other than Visual Studio / VSCode / Xamarin etc. 

As for cmake-tools, I recall that I built it without cpp tools and it worked fine. I think the reason why cmake-tools depends on cpp-tools is because cmake-tools updates the intellisense database. Still, configuring and building projects "should" work.

spoenemann:
Ok, could you remove cpptools, then we'll merge cmake-tools?

kitattyor:
Done .. Sorry for the delay..

spoenemann:
Thanks! The remaining question is: shall we refer to a specific version and tag to ensure we publish only the officially released versions? In that case we would also need `version` and `checkout` properties as described in the README here.

An alternative is to just download a `vsix` file from the GitHub releases. I have seen the `vsix` packages are attached to the releases in the cmake-tools repository. In that case you can use the `download` property instead of `repository`.

spoenemann:
So here are the two alternatives:
```json
    {
      "id": "microsoft.vscode-cmake-tools",
      "repository": "https://github.com/microsoft/vscode-cmake-tools",
      "version": "1.4.2",
      "checkout": "1.4.2"
    },
```
or
```json
    {
      "id": "microsoft.vscode-cmake-tools",
      "version": "1.4.2",
      "download": "https://github.com/microsoft/vscode-cmake-tools/releases/download/1.4.2/cmake-tools.vsix"
    },
```

spoenemann:
The download variant is safer because we don't risk to mess up the package by doing something wrong in the build process.

kitattyor:
Done :)

spoenemann (approved):
Thanks!

PR #181, opened 2020-10-01
Add spring-initializr

herandol:
Add spring-initializr vscode plugin.

### DIFF
--- a/extensions.json
+++ b/extensions.json
@@ -1087,6 +1087,12 @@
       "prepublish": "npm run build-plugin"
     },
     {
+      "id": "vscjava.vscode-spring-initializr",
+      "repository": "https://github.com/microsoft/vscode-spring-initializr",
+      "version": "0.5.0",
+      "prepublish": "npm run vscode:prepublish"
+    },
+    {
       "id": "vscode-icons-team.vscode-icons",
       "repository": "https://github.com/vscode-icons/vscode-icons",
       "version": "11.0.0",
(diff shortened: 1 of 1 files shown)


extensions.json
```diff
@@ -1086,6 +1086,12 @@
       "checkout": "0.24.2",
       "prepublish": "npm run build-plugin"
     },
+    {
+      "id": "vscjava.vscode-spring-initializr",
+      "repository": "https://github.com/microsoft/vscode-spring-initializr",
+      "version": "0.5.0",
```
jankeromnes, line 1092: Nit: Using `"version"` without a corresponding `"checkout"` Git SHA will publish once from the default branch (even if different version).

But the default branch may contain unrelated changes (or even a different version altogether), so I think adding a `"checkout"` parameter is safer:

```suggestion
      "version": "0.5.0",
      "checkout": "v0.5.0"
```

extensions.json
```diff
@@ -1086,6 +1086,12 @@
...
+      "prepublish": "npm run vscode:prepublish"
```
jankeromnes, line 1093: Nit: `ovsx publish` will already run `vscode:prepublish` (because it's based on `vsce`).

The optional `"prepublish"` parameter here is just so that you can provide additional steps that extensions may not have included in a `vscode:prepublish` already (e.g. some extensions need custom steps like a `npm run compile` or `yarn build` or `yarn package`).

Here, I think we won't need that.

```suggestion
```

jankeromnes (commented):
Many thanks for the contribution!

Two small things to adjust, but otherwise this looks good to me.

Also I'll be able to automatically apply my own suggestions & merge this without further bothering you, so I'll do that. 🙂

jankeromnes:
FYI, publishing was successful! 🎉 https://open-vsx.org/extension/vscjava/vscode-spring-initializr

PR #186, opened 2020-10-06
Publish ms-vscode.node-debug2 extension

shahar-h:


jankeromnes:
Publishing worked! 🎉 https://open-vsx.org/extension/ms-vscode/node-debug2

PR #189, opened 2020-10-13
Add Nix language support

1000101:
Add Nix language support as it's currently missing in OVSX registry - https://github.com/bbenoist/vscode-nix/issues/7

### DIFF
--- a/extensions.json
+++ b/extensions.json
@@ -128,6 +128,12 @@
       "checkout": "1.0.0"
     },
     {
+      "id": "bbenoist.Nix",
+      "repository": "https://github.com/bbenoist/vscode-nix",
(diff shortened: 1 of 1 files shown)


spoenemann (approved):
Thanks. However, we should remove this entry again in case @bbenoist decides to publish the extension.

1000101:
I've just found out that there is a better, up-to-date initiative in https://github.com/nix-community/vscode-nix-ide/issues/17/ so once it gets published this PR can be reverted in its favor. Sorry for the fuss!!!

spoenemann:
Ok I'll revert the commit.

1000101:
Thank you so much!

PR #190, opened 2020-10-14
Add local-lua-debugger-vscode extension.

eduardomezencio:


extensions.json
```diff
@@ -1096,6 +1096,11 @@
       "version": "0.10.1",
       "checkout": "v0.10.1"
     },
+    {
+      "id": "tomblind.local-lua-debugger-vscode",
+      "repository": "https://github.com/tomblind/local-lua-debugger-vscode",
+      "version": "0.2.0"
```
spoenemann, line 1102:
```suggestion
      "version": "0.2.0",
      "checkout": "release-0.2.0",
      "prepublish": "npm run build"
```

spoenemann (changes requested):
Thanks!

I suggest checking out the release branch to ensure we build from the right version, and to add the build script (it is not run automatically in this case).

extensions.json
eduardomezencio: Thanks, I'll get it fixed. Pardon my ignorance, but how exactly would I know if for a given package this prepublish step is required? And would not the npm run publish also be needed? I just would like to have this clear on my mind so that I won't repeat mistakes.

eduardomezencio:
I commented after your code review, @spoenemann , but just repeating here in case you don't see it. How can I know exactly if an extension needs an additional step, like `npm run build` in this case, between the `npm install` and `ovsx publish`?

spoenemann:
Usually one can see that such an additional step is required by looking at the `scripts` section of the extension's package.json.

spoenemann (approved):

PR #198, opened 2020-10-23
Add ctf0.macros

tywmick:
[macros](https://github.com/ctf0/macros) is an MIT-licensed extension that adds support for custom macros.

```bash
vsce package
code --install-extension macros-0.0.4.vsix
```

...is all that's needed to install the extension. 👍🏼


spoenemann (approved):
Thanks!

PR #206, opened 2020-11-05
Fix syntax error in extensions.json

shahar-h:
Currently publish extensions process is broken because of syntax error in extensions.json:
image


shahar-h:
@akosyakov can you please review?

PR #202, opened 2020-11-01
Add Unity related extension

Awea:
Hi there,

This pull request:
- Fix a JSON error in extensions.json
- Add https://github.com/TobiahZ/unity-tools (under MIT license)
- Add https://github.com/Unity-Technologies/vscode-unity-debug (under MIT license)

PR #215, opened 2020-11-12
Add Unity.unity-debug using .vsix file

Awea:
Related to https://github.com/open-vsx/publish-extensions/pull/202#issuecomment-726019395

### DIFF
--- a/extensions.json
+++ b/extensions.json
@@ -1146,9 +1146,8 @@
     },
     {
       "id": "Unity.unity-debug",
-      "repository": "https://github.com/Unity-Technologies/vscode-unity-debug",
-      "version": "3.0.0",
-      "checkout": "Version-3.0.1"
+      "download": "https://github.com/Unity-Technologies/vscode-unity-debug/releases/download/Version-2.7.2/unity-debug-2.7.2.vsix",
+      "version": "2.7.2"
     },
     {
       "id": "vhdlwhiz.vhdl-by-vhdlwhiz",


jankeromnes (approved):
Looks good to me, many thanks! 🚀

jankeromnes:
This should successfully publish `Unity.unity-debug` tonight at around [03:03 UTC](https://github.com/open-vsx/publish-extensions/blob/b02685d6406c9d5b36676257979a2834ee92e76e/.github/workflows/publish-extensions.yml#L3-L6).

jankeromnes:
@Awea Oh no, auto-publishing failed again, this time with:

```
Downloading https://github.com/Unity-Technologies/vscode-unity-debug/releases/download/Version-2.7.2/unity-debug-2.7.2.vsix
  "id": "Unity.unity-debug",
  "download": "https://github.com/Unity-Technologies/vscode-unity-debug/releases/download/Version-2.7.2/unity-debug-2.7.2.vsix",
  "version": "2.7.2"
}
Error: Cannot resolve dependency: ms-vscode.csharp
```

And unfortunately, `ms-vscode.csharp` cannot be uploaded to OpenVSX because [its LICENSE](https://marketplace.visualstudio.com/items/ms-dotnettools.csharp/license) forbids it. 😕

Awea:
index

@jankeromnes I'm sad to read this, should I remove this extension from the extensions.json file and forget about it?

jankeromnes:
@Awea I'm not exactly sure. It's not a big issue to keep a failing extension in the list (the CI handles it pretty well) and I really hope that we'll be able to publish it some day. I don't like that non-Microsoft tools like VSCodium, Gitpod, Theia, Eclipse Che, etc cannot use some really cool open-source extensions like `Unity.unity-debug` just because of a bad dependency license.

One interesting question is how strongly tied to `ms-vscode.csharp` the `Unity.unity-debug` extension is, and whether it could also work with a different, not-Microsoft-only alternative (e.g. something like [free-omnisharp-vscode](https://github.com/muhammadsammy/free-omnisharp-vscode) maybe?)

Note that there is a similar discussion going on in https://github.com/platformio/platformio-vscode-ide/issues/1802 where the PlatformIO extension depends on `ms-vscode.cpptools` (which has a Microsoft-only license) but only for 5% of its features, so they're currently considering alternative dependencies to fix the license problem.

Awea:
Hi @jankeromnes, thanks for the details. I neither don't like the fact that VSCodium can't use this extension because of some license issue. 

I'll try to ask on Unity.unity-debug if the extension can use a free alternative for csharp like you suggested :)

PR #213, opened 2020-11-11
Add Erlang LS extension

chintan-mishra:
Add plugins for supporting Erlang

PR #217, opened 2020-11-13
Amend 'Add Erlang LS extension (#213)'

chintan-mishra:


chintan-mishra:
The name of `ErlangLS` plugin are of the following format: `erlang-ls-0.0.15.vsix`. I have directed the auto-publishing tool to use the wildcard filename whenever a new `.vsix` file is built. In order to do this, I have used `extensionFile: erlang-ls-*.vsix` in the section defining this tool. What are the possible ramifications of this step?

extensions.json
```diff
@@ -1307,7 +1307,10 @@
     },
     {
         "id": "erlang-ls.erlang-ls",
-        "repository": "https://github.com/erlang-ls/vscode"
+        "repository": "https://github.com/erlang-ls/vscode",
+        "checkout": "master",
+        "prepublish": "npm run compile",
```
jankeromnes, line 1312: I notice that the [vscode:prepublish](https://github.com/erlang-ls/vscode/blob/79b7b5b8a180f00c640d9aa772e2fc4d9d7180e7/package.json#L22) NPM script already runs `cd client && npm run compile`, but I agree that the [compile](https://github.com/erlang-ls/vscode/blob/79b7b5b8a180f00c640d9aa772e2fc4d9d7180e7/package.json#L23) NPM script is more comprehensive as it also runs:

```
cd erlang_ls && rebar3 as debug escriptize
```

So, okay for adding `npm run compile` as an extra `prepublish` step here.

extensions.json
```diff
@@ -1307,7 +1307,10 @@
...
+        "extensionFile": "erlang-ls-*.vsix"
```
jankeromnes, line 1313: I don't think that the `extensionFile` attribute here supports wildcards (`*`).

But why do you feel like providing `extensionFile` is needed? (`ovsx` already knows how to package and publish a `.vsix` file without extra guidance)

jankeromnes (commented):
Thanks for improving the auto-publishing of the Erlang LS extension!

I've added a few thoughts & questions below.

extensions.json
```diff
@@ -1307,7 +1307,10 @@
     },
     {
         "id": "erlang-ls.erlang-ls",
-        "repository": "https://github.com/erlang-ls/vscode"
+        "repository": "https://github.com/erlang-ls/vscode",
+        "checkout": "master",
```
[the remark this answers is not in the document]
chintan-mishra: Acknowledged. I will delete this line in the next commit after resolving all other concerns.

extensions.json
chintan-mishra: Acknowledged

extensions.json
chintan-mishra: When using `vsce package`, the `*.vsix` file is placed in the root folder of the repository.

After building the `*.vsix` file, where does the `osvx` look for this file?

> But why do you feel like providing extensionFile is needed?

Based on some other pull requests on this repository, I noticed we have to update version number after each new release. So, I thought of reducing technical debt with the use of a wildcard.

If `osvx` can find the published extension, then we don't need extra declarations that will otherwise create confusion.

```suggestion

```

jankeromnes: Agreed, without the `"extensionFile"` should work fine (actually it already worked once without it: https://open-vsx.org/extension/erlang-ls/erlang-ls)

chintan-mishra: Makes sense.

extensions.json
chintan-mishra: Resolved in the latest commit. Please verify.

extensions.json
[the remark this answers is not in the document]
chintan-mishra: Resolved in the latest commit. Please verify.

jankeromnes (approved):
Looks good to me, many thanks! 💯 🙏

PR #222, opened 2020-11-17
Download VSIX for OCaml Platform extension

mnxn:
The 1.4.0 version of the OCaml Platform extension is currently broken when downloaded from open-vsx:
https://github.com/eclipse/openvsx/issues/203
https://github.com/ocamllabs/vscode-ocaml-platform/issues/436

The extension has some specific build requirements that would prevent it from building correctly with the current configuration. This PR switches to downloading a released VSIX that is already built.

### DIFF
--- a/extensions.json
+++ b/extensions.json
@@ -922,9 +922,8 @@
     },
     {
       "id": "ocamllabs.ocaml-platform",
-      "repository": "https://github.com/ocamllabs/vscode-ocaml-platform.git",
-      "version": "1.4.0",
-      "checkout": "v1.4.0"
+      "download": "https://github.com/ocamllabs/vscode-ocaml-platform/releases/download/v1.4.0/ocaml-platform-1.4.0.vsix",
+      "version": "1.4.0"
     },
     {
       "id": "octref.vetur",


spoenemann:
Hi @mnxn, thanks for the contribution! The change looks good. But an even better alternative would be if you could add a publishing step in the source repository. The CLI is named [ovsx](https://www.npmjs.com/package/ovsx) and is very similar to `vsce`, which you are already using in the [package.json](https://github.com/ocamllabs/vscode-ocaml-platform/blob/cac6fa025487a2fb209632c58ef3a6ad4508e2f9/package.json#L573). See [this Wiki page](https://github.com/eclipse/openvsx/wiki/Publishing-Extensions) for more details. The difference would be that the extension would be shown as "Published by @mnxn" with a "verified" icon (like [Julia](https://open-vsx.org/extension/julialang/language-julia), for example), while when merging this PR the extension continues to be shown as "Published by @open-vsx" with a "public" icon (see [OCaml Platform](https://open-vsx.org/extension/ocamllabs/ocaml-platform)).

mnxn:
Hi, thanks for the instructions.

I tried to publish the extension but I get the error that
```
Extension ocamllabs.ocaml-platform version 1.4.0 is already published.
```
Reading through the [`publish-extensions` script](https://github.com/open-vsx/publish-extensions/blob/e70fb554a5c265e53f44605dbd826270b860694b/publish-extensions.js#L49), it seems that merging this PR would have no effect too since the version number won't change.

The problem is that the VSIX published to open-vsx is broken and incomplete when compared to the vscode marketplace one. In the future we'll try to add a CI/CD step the publishes to open-vsx, but the next release might not be for a while.

Is there any way to replace the currently published 1.4.0 version?

spoenemann:
I deleted version `1.4.0` and removed the entry in `extensions.json` so you are now free to publish it yourself.

Are the earlier versions affected, too? We currently have `0.9.0`, `1.1.1` and `1.2.0`.

mnxn:
Thank, I published it!

Versions `0.9.0`, `1.1.1`, and `1.2.0` are not affected.

PR #240, opened 2020-12-01
Add vscode-syncing

brianclinkenbeard:
Licensed under MIT. Added with node script.

### DIFF
--- a/extensions.json
+++ b/extensions.json
@@ -943,7 +943,8 @@
     {
       "id": "nonoroazoro.syncing",
       "repository": "https://github.com/nonoroazoro/vscode-syncing",
-      "version": "3.1.0"
+      "version": "3.1.0",
+      "checkout": "v3.1.0"
     },
     {
       "id": "notskm.clang-tidy",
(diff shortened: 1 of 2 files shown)


extensions.json
```diff
@@ -940,6 +940,11 @@
       "id": "Nightrains.robloxlsp",
       "repository": "https://github.com/NightrainsRbx/RobloxLsp"
     },
+    {
+      "id": "nonoroazoro.syncing",
+      "repository": "https://github.com/nonoroazoro/vscode-syncing",
+      "version": "3.1.0"
```
jankeromnes, line 946: Since we're pinning a specific version, we should probably publish from the corresponding release tag (instead of the repository's default branch, which may contain WIP changes at the time of publishing):

```suggestion
      "version": "3.1.0",
      "checkout": "v3.1.0"
```

jankeromnes (approved):
Great, thanks! 👍 

One small adjustment needed, but I can apply it on GitHub before merging. 🚀

PR #251, opened 2020-12-27
Add arcanis.vscode-zipfs

pschyska:
This adds arcanis.vscode-zipfs.
The License is [2-Clause-BSD](https://github.com/yarnpkg/berry/blob/master/LICENSE.md).
Upstream doesn't seem to tag, so I didn't add a checkout.
There is an Issue for them to publish themselves, but there wasn't progress so far. Can the yarn project claim the namespace arcanis later, when it's added through here?

### DIFF
--- a/extensions.json
+++ b/extensions.json
@@ -95,9 +95,10 @@
     },
     {
       "id": "arcanis.vscode-zipfs",
-      "repository": "https://github.com/yarnpkg/berry.git",
+      "repository": "https://github.com/yarnpkg/berry",
       "version": "2.2.4",
-      "location": "packages/vscode-zipfs/"
+      "checkout": "vscode-zipfs/2.2.4",
+      "location": "packages/vscode-zipfs"
     },
     {
       "id": "arjun.swagger-viewer",
--- a/extensions.json
+++ b/extensions.json
@@ -94,6 +94,12 @@
       "version": "1.0.3"
     },
     {
+      "id": "arcanis.vscode-zipfs",
+      "repository": "https://github.com/yarnpkg/berry.git",
+      "version": "2.2.4",
+      "location": "packages/vscode-zipfs/"
+    },
+    {
       "id": "arjun.swagger-viewer",
       "repository": "https://github.com/arjun-g/vs-swagger-viewer"
     },


extensions.json
```diff
@@ -93,6 +93,12 @@
       "repository": "https://github.com/appulate/vscode-file-watcher",
       "version": "1.0.3"
     },
+    {
+      "id": "arcanis.vscode-zipfs",
+      "repository": "https://github.com/yarnpkg/berry.git",
+      "version": "2.2.4",
```
jankeromnes, line 99: When pinning a specific version, it's better to checkout the corresponding release tag or branch (as opposed to publishing straight from the default branch, which may contain unwanted/unfinished changes):

```suggestion
      "version": "2.2.4",
      "checkout": "vscode-zipfs/2.2.4",
```

extensions.json
```diff
@@ -93,6 +93,12 @@
       "repository": "https://github.com/appulate/vscode-file-watcher",
       "version": "1.0.3"
     },
+    {
+      "id": "arcanis.vscode-zipfs",
+      "repository": "https://github.com/yarnpkg/berry.git",
```
jankeromnes, line 98: While this should work as is, I personally prefer repository URLs without the unnecessary `.git` suffix:

```suggestion
      "repository": "https://github.com/yarnpkg/berry",
```

extensions.json
```diff
@@ -93,6 +93,12 @@
...
+      "version": "2.2.4",
+      "location": "packages/vscode-zipfs/"
```
jankeromnes, line 100: No need for the final slash:

```suggestion
      "location": "packages/vscode-zipfs"
```

jankeromnes (approved):
Many thanks for your contribution @pschyska!

This change mostly looks good to me, with a few minor adjustments that I'll be able to apply before merging (so there is nothing else you need to do here).

Happy last day of 2020! 🎆

jankeromnes:
Ah, I forgot to reply to your comments/questions:

> Upstream doesn't seem to tag, so I didn't add a checkout.

I found a suitable-looking release tag (using an unconventional format). 😇

> There is an Issue for them to publish themselves, but there wasn't progress so far. Can the yarn project claim the namespace arcanis later, when it's added through here?

Sure, the owners/maintainers of `arcanis.vscode-zipfs` (hi @arcanis! 👋) are more than welcome to claim the `arcanis` namespace on https://open-vsx.org at any time. Please ping me if I can help! 👍

pschyska:
> Ah, I forgot to reply to your comments/questions:
> 
> > Upstream doesn't seem to tag, so I didn't add a checkout.
> 
> I found a suitable-looking release tag (using an unconventional format). 

Thanks, I must be blind!

pschyska:
> Happy last day of 2020!

Happy new year!

jankeromnes:
Thanks @pschyska!

Unfortunately, auto-publishing this extension failed with:

```
  "id": "arcanis.vscode-zipfs",
  "repository": "https://github.com/yarnpkg/berry",
  "version": "2.2.4",
  "checkout": "vscode-zipfs/2.2.4",
  "location": "packages/vscode-zipfs"
}
Error: Invalid URL in field 'repository': ssh://git@github.com/yarnpkg/berry.git
```

To fix this, the `repository` URL in the extension's `package.json` needs to use the supported format: `https://github.com/yarnpkg/berry`.

Or, alternatively, you could also opt to hot-patch the extension's `package.json` using a custom `prepublish` script, e.g. a bit like so:

https://github.com/open-vsx/publish-extensions/blob/aea1532d7a4eb8a31ec8fb1949fd0684897c74fd/extensions.json#L371

spoenemann:
Maybe we should update the validation to support ssh urls? See https://github.com/eclipse/openvsx/issues/128

pschyska:
@jankeromnes That's unfortunate. I second @spoenemann 's opinion that ssh URLs should be valid though. `ssh://git@github.com/yarnpkg/berry.git` is perfectly fine IMHO (I know that in the Github-era ssh can be problematic, but I prefer the protocol anyways because it allows me to use a hardware token for auth).
In the meantime, patching the package.json should be fine, I would try to maybe find a sed call with higher specifivity though.
I can look into it soon-ish.

pschyska:
@jankeromnes I'm noticing some internal resistance to sed the `package.json` somehow.
Wouldn't it make more sense to add a node script like the following to the repo:
````normalize-git-urls.js
const package = require("./package.json");
const { writeFileSync } = require("fs");
package?.repository?.url.replace(/git\+https/, "https");
writeFileSync("foo.json", JSON.stringify(p, null, 2) + "\n");
````
This would have the advantage of not accidentally replacing text somewhere else in the `package.json`, not depending on sed availability in CI, and would make it easier to also transform `git@github.com...` into `https://github.com...` (but I'm not sure that's something you'd want?).

jankeromnes:
@pschyska Interesting suggestion, thanks!

However, I'm personally leaning against this, because:

- typically extensions that require extra special processing in their auto-publishing process are pretty rare (only 22 in the 294 [extensions.json](https://github.com/open-vsx/publish-extensions/blob/master/extensions.json) extensions are "special" and need a custom `"prepublish"` step)

- also, their "fix" is usually pretty unique (e.g. running some extra build command that was not added in the standard `vscode:prepublish` step, or installing some extra missing tool or dependency, or fixing the odd unusual URL format in `package.json`, etc). I don't think we can or should modify our publishing CI every time we encounter something unusual in an extension, and instead I'd argue that's what the `"prepublish"` step is for

(E.g. if you search for `"prepublish"` in [extensions.json](https://github.com/open-vsx/publish-extensions/blob/master/extensions.json), you'll see how varied all these required custom extra steps are.)

Also, it's worth noting that https://github.com/open-vsx/publish-extensions is kind of an interim solution. In the ideal scenario, every extension owner can run `npx ovsx publish` in the same script or CI job that also publishes to VSCode Marketplace, but sometimes this cannot be achieved quickly and thus we add the extension here as a shortcut.

pschyska:
The real fix would be for Java to accept the Scheme `git+https`, which is and always was valid AFAIK :-)
So let's use the sed workaround.

pschyska:
Added #261 with your suggested workaround.

PR #255, opened 2021-01-01
Add prepublish script for vsc-material-theme and vsc-material-theme-icons

StevenDoesStuffs:
Both vsc-material-theme and vsc-material-theme-icons require `yarn build` to be run before publishing, and this is currently not being done. 

The build script for vsc-material-theme-icons clones the repository material-theme/vsc-material-theme-icons-src. Unfortunately, it tries to clone the repo through ssh, which requires a public key to be present. I'm using sed to make the build script clone the https version of the same repo. 

Should resolve https://github.com/material-theme/vsc-material-theme-icons/issues/158, https://github.com/material-theme/vsc-material-theme/issues/496, and https://github.com/material-theme/vsc-material-theme/issues/507.

### DIFF
--- a/extensions.json
+++ b/extensions.json
@@ -376,11 +376,13 @@
     },
     {
       "id": "Equinusocio.vsc-material-theme",
-      "repository": "https://github.com/material-theme/vsc-material-theme"
+      "repository": "https://github.com/material-theme/vsc-material-theme",
+      "prepublish": "yarn build"
     },
     {
       "id": "equinusocio.vsc-material-theme-icons",
-      "repository": "https://github.com/material-theme/vsc-material-theme-icons"
+      "repository": "https://github.com/material-theme/vsc-material-theme-icons",
+      "prepublish": "sed -i 's+git@github.com:material-theme/vsc-material-theme-icons-src.git+https://github.com/material-theme/vsc-material-theme-icons-src.git+g' build/get-remote-icons.ts && yarn build"
     },
     {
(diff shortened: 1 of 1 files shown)


jankeromnes (approved):
Wow, many thanks for tracking down these problems & fixing them with `prepublish` steps! 🚀

On a side note, the same could be achieved upstream with a `vscode:prepublish` command in `package.json`. But since this fix looks correct, let's merge & deploy it.

I guess we'll also need to delete all previous versions of both extensions from https://open-vsx.org once fixed versions have been uploaded.

jankeromnes:
@spoenemann Could you please delete the latest published version of both https://open-vsx.org/extension/Equinusocio/vsc-material-theme and https://open-vsx.org/extension/Equinusocio/vsc-material-theme-icons ?

This is to allow fixed binaries to be re-uploaded tonight.

spoenemann:
Done. Are the other versions correct?

jankeromnes:
@spoenemann thanks! No, unfortunately all current versions are broken, but I was waiting for a fixed version to be uploaded before requesting that all others be deleted.

jankeromnes:
@spoenemann Ok, the latest published versions (`33.1.2` and `1.2.2`) should now be fixed. Could you please remove all older versions?

Also, @StevenDoesStuffs could you please confirm that the latest published versions now work as expected?

StevenDoesStuffs:
It appears to be working.

spoenemann:
Removed older versions.

PR #261, opened 2021-01-13
Patch repo URL in package.json, as git+https is not allowed

pschyska:
This is a follow-up to #251, as it was already merged.
The server doesn't accept the repo URL in the `package.json`, as it uses the `git+https` scheme. The prepublish step rewrites that to `https`, which is also valid.

### DIFF
--- a/extensions.json
+++ b/extensions.json
@@ -108,7 +108,8 @@
       "repository": "https://github.com/yarnpkg/berry",
       "version": "2.2.4",
       "checkout": "vscode-zipfs/2.2.4",
-      "location": "packages/vscode-zipfs"
+      "location": "packages/vscode-zipfs",
+      "prepublish": "sed -ri \"s_git\\+https://_https://_g\" package.json"
     },
     {
       "id": "arjun.swagger-viewer",
--- a/extensions.json
+++ b/extensions.json
@@ -109,7 +109,7 @@
       "version": "2.2.4",
       "checkout": "vscode-zipfs/2.2.4",
       "location": "packages/vscode-zipfs",
-      "prepublish": "sed -ri \"s_git\\+https://_https://_g\" package.json"
+      "prepublish": "sed -ri \"s_ssh://git@github.com/yarnpkg/berry.git_https://github.com/yarnpkg/berry_g\" package.json"
     },
     {
       "id": "arjun.swagger-viewer",


extensions.json
```diff
@@ -108,7 +108,8 @@
       "repository": "https://github.com/yarnpkg/berry",
       "version": "2.2.4",
       "checkout": "vscode-zipfs/2.2.4",
-      "location": "packages/vscode-zipfs"
+      "location": "packages/vscode-zipfs",
+      "prepublish": "sed -ri \"s_git\\+https://_https://_g\" package.json"
```
jankeromnes, line 112: Hmm, looking back at the auto-publishing error, it says:

```
Error: Invalid URL in field 'repository': ssh://git@github.com/yarnpkg/berry.git
```

So I don't think this has to do with `git+https://` vs `https://`, but about the `ssh://git@github.com`.

Instead, maybe it's better to replace the entire offending URL with an acceptable variant (fewer risks of accidental side effects when the pattern is more specific):

```suggestion
      "prepublish": "sed -ri \"s_ssh://git@github.com/yarnpkg/berry.git_https://github.com/yarnpkg/berry_g\" package.json"
```

jankeromnes (commented):
Many thanks @pschyska for accepting this approach and implementing a `package.json` hotfix `"prepublish"` step!

Approved but with one change request. Please take a look and let me know if you accept the suggestion (if you do, I can automatically commit the suggestion from GitHub before merging, so you don't actually need to make a change to this PR).

extensions.json
pschyska: lgtm,

jankeromnes (approved):

jankeromnes:
Oops, still fails:

```
  "id": "arcanis.vscode-zipfs",
  "repository": "https://github.com/yarnpkg/berry",
  "version": "2.2.4",
  "checkout": "vscode-zipfs/2.2.4",
  "location": "packages/vscode-zipfs",
  "prepublish": "sed -ri \"s_ssh://git@github.com/yarnpkg/berry.git_https://github.com/yarnpkg/berry_g\" package.json"
}
Error: Invalid URL in field 'repository': ssh://git@github.com/yarnpkg/berry.git
```

Actually I think the replacement needs to be made to `packages/vscode-zipfs/package.json`, not to `package.json`.

Will make a commit.

PR #263, opened 2021-01-14
Add jebbs.plantuml

marcdumais-work:
By trial-and-error I came to the conclusion that this extension requires one to
add an official "plantuml.jar" in the extension's root, that will be used for
local, serverless rendering. This apparently does not happen automatically.

That plantuml.jar is available under a few different licenses. I chosed the MIT
one, to match this extension's license. When the zip file is extracted a "COPYING"
file is created along with plantuml.jar, that documents its provenamce and license.
For clarity I rename this file to COPYING.plantuml-jar and leave it in place to be
packaged along with the extension.

Here's where the URL used to fetch plantuml.jar comes from:

Start at the official PlantUML download page:
https://plantuml.com/download
Chose the MIT-version at the bottom of the page and copy the sourceforge link.
At the time of writing it is:
http://sourceforge.net/projects/plantuml/files/plantuml-jar-mit-1.2021.0.zip/download

Signed-off-by: Marc Dumais <marc.dumais@ericsson.com>

### DIFF
--- a/extensions.json
+++ b/extensions.json
@@ -665,6 +665,13 @@
       "version": "1.2.3"
     },
     {
+      "id": "jebbs.plantuml",
+      "version": "2.14.0",
+      "checkout": "v2.14.0",
+      "repository": "https://github.com/qjebbs/vscode-plantuml.git",
+      "prepublish": "wget -O plantuml.zip https://sourceforge.net/projects/plantuml/files/1.2021.0/plantuml-jar-mit-1.2021.0.zip/download && unzip plantuml.zip && rm plantuml.zip && mv COPYING COPYING.plantuml-jar"
+    },
+    {
       "id": "jock.svg",
       "repository": "https://github.com/lishu/vscode-svg2",
       "version": "1.4.3"


extensions.json
```diff
@@ -664,6 +664,12 @@
       "repository": "https://github.com/JeffersonQin/VSCode-LaTeX-Snippets",
       "version": "1.2.3"
     },
+    {
+      "id": "jebbs.plantuml",
+      "version": "2.13.12",
```
marcdumais-work, line 669: I am not sure if this field does what I hope it does. I hope it would tell the build service to start with that version of the extension, and then continue publishing more recent ones as well?

spoenemann: By declaring the version, the script checks whether the version in the package.json matches that version. It is not updated automatically.

The latest version seems to be 2.14.1 (`master` branch), but the latest tag is `v2.14.0`. I'd propose to follow the tags and add `"checkout": "v2.14.0"`, and to adapt the version accordingly.

jankeromnes: > By declaring the version, the script checks whether the version in the package.json matches that version. It is not updated automatically.

Slight precision here: `extension.version` is only compared against what is already on OpenVSX: https://github.com/open-vsx/publish-extensions/blob/994cbe51fcabac973de54445f15344804a6f8cd5/publish-extensions.js#L65-L74

- if `extension.version` is **older** than what's on OpenVSX, the script throws an error (i.e. we're trying to publish something older than what's on OpenVSX, which likely means someone else took over the publishing & we should stop doing that here)
- if `extension.version` is **equal** to what's on OpenVSX, this extension is simply skipped (no need to re-publish the same version again)
- if `extension.version` is **newer** than what's on OpenVSX (or if there is nothing on OpenVSX, or if `extension.version` isn't specified), we go ahead and try to publish

So the version specified in the extension's `package.json` is never checked by our script (however, if the actual `package.json` version is equal or older to what's on OpenVSX, publishing will still fail but at a later point, i.e. in `ovsx` CLI itself)

spoenemann: Thanks for the clarification!

> if the actual package.json version is equal or older to what's on OpenVSX, publishing will still fail but at a later point, i.e. in ovsx CLI itself

Slight precision here: ovsx allows to publish older version. It fails only if the exact version is already published.

marcdumais-work: > I'd propose to follow the tags and add "checkout": "v2.14.0", and to adapt the version accordingly.

Done.

marcdumais-work: I think it would be awesome to have a way to request the build service to build a "one-off", missing version of an extension it already builds. For many reasons it can be desirable to use a specific, older version, of an extension. 

Maybe this could be done by submitting a PR to an alternate branch, that would upon merge trigger the build service for whatever extension(s) have changed?

marcdumais-work:
@jankeromnes  @spoenemann  I think I did the requested modifications. Can you have another look?

spoenemann (approved):
Thanks, let's try it!

marcdumais-work:
I do not see the expected version on `open-vsx.org`. Is there a way to know if/why the build failed (I assume)?

marcdumais-work:
Trying to built the extension locally does not work for me with v2.14.1. v2.13.15 does. I'll do a quick PR to try with that version instead.

marcdumais-work:
I get the same errors with latest master, too. Will need to troubleshoot - I assume the publish bot is encountering the same.

jankeromnes:
> I do not see the expected version on `open-vsx.org`. Is there a way to know if/why the build failed (I assume)?

This repository uses a GitHub Action to run the publishing script.

You can check the logs by going into:
1. [Actions](https://github.com/open-vsx/publish-extensions/actions)
2. Select the latest workflow called [Publish extensions to open-vsx.org](https://github.com/open-vsx/publish-extensions/actions/runs/526543027)
3. Then click on the job called [node publish-extensions](https://github.com/open-vsx/publish-extensions/runs/1803737699?check_suite_focus=true) to see the logs (and e.g. search for your extension's ID)

In your extension's case, the latest run says:

```
Checking Open VSX version of jebbs.plantuml
Found version: 2.14.0
[SKIPPED] Requested version 2.14.0 is already published on Open VSX
```

So it's just skipping the publish step because the specified `"version"` is already on Open VSX (if it's broken, we'd need to delete it before it can be re-published).

marcdumais-work:
Thanks @jankeromnes . I think publishing worked since that version was published by `open-vsx` 6 days ago. So it's not re-publishing it needlessly, which is good!

image


image

marcdumais-work:
> You can check the logs by going into:

Thanks! I'll know how to do it next time.

PR #265, opened 2021-01-15
Added workflow to validate extensions.json

spoenemann:
This will run on every push and pull request to validate that extensions.json can be parsed and does not contain invalid entries. We could consider adding more in-depth validations later.


spoenemann:
Merging now, feel free to change the code later.

PR #267, opened 2021-01-18
feat: Add oguhpereira.vscode-spotify-theme

vymarkov:
LICENSE is present
https://github.com/oguhpereira/vscode-spotify-theme/blob/master/LICENSE

### DIFF
--- a/extensions.json
+++ b/extensions.json
@@ -75,15 +75,15 @@
       "repository": "https://github.com/almenon/arepl-vscode"
     },
     {
+      "id": "amazonwebservices.aws-toolkit-vscode",
+      "download": "https://github.com/aws/aws-toolkit-vscode/releases/download/v1.18.0/aws-toolkit-vscode-1.18.0.vsix",
+      "version": "1.18.0"
+    },
+    {
       "id": "amiralizadeh9480.laravel-extra-intellisense",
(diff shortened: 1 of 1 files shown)


extensions.json
```diff
@@ -1147,6 +1147,11 @@
       "id": "oderwat.indent-rainbow",
       "repository": "https://github.com/oderwat/vscode-indent-rainbow"
     },
+    {
+      "id": "oguhpereira.spotify-color-theme",
+      "repository": "https://github.com/oguhpereira/vscode-spotify-theme",
+      "version": "1.2.9"
```
spoenemann, line 1153:
```suggestion
      "version": "1.2.9",
      "checkout": "v1.2.9"
```

spoenemann (commented):
Thanks! When a tag is available for the version to be installed, I think we should use it.

extensions.json
vymarkov: > Thanks! When a tag is available for the version to be installed, I think we should use it.

@spoenemann I just wanna note that I prefer to use `npm run add`, it requires less time to add a new extension, I added this one by using command below:
```bash
$  npm run add -- https://github.com/oguhpereira/vscode-spotify-theme
```

It looks like we can improve `add-extension.js` util to find the corresponding tag and use it instead of doing manual steps for this.

What do you think about it?

PR #270, opened 2021-01-20
feat(add-extension): add bierner.markdown-image-size

joaosa:
Adding https://github.com/mjbvz/vscode-markdown-image-size. It has an MIT
licence.

### DIFF
--- a/extensions.json
+++ b/extensions.json
@@ -897,6 +897,11 @@
       "version": "1.4.0"
     },
     {
+      "id": "bierner.markdown-image-size",
+      "repository": "https://github.com/mjbvz/vscode-markdown-image-size",
+      "version": "0.0.4"
+    },
+    {
       "id": "mp.oradew-vscode",
       "repository": "https://github.com/mickeypearce/oradew-vscode",
       "version": "0.3.27",


extensions.json
```diff
@@ -896,6 +896,10 @@
       "repository": "https://github.com/mitchdenny/ecdc",
       "version": "1.4.0"
     },
+    {
+      "id": "mjbvz.vscode-markdown-image-size",
```
spoenemann, line 900: The extension has a different package and publisher name in its [package.json](https://github.com/mjbvz/vscode-markdown-image-size/blob/master/package.json): `bierner.markdown-image-size`

joaosa: My bad! This should be ok now

spoenemann (approved):
Thank you!

joaosa:
Hey! It doesn't seem like this is getting published. Could it be an issue with the specified version (0.0.4) not being tagged on the repo?

spoenemann:
Another extension seemed to stop the publishing workflow because the CLI expected user input. I removed that extension, let's see again tomorrow.

PR #271, opened 2021-01-20
Add Ionide F# extension

bryphe:
Hi,

Would it be possible to add the `Ionide.Ionide-Fsharp` extension? License is [MIT](https://github.com/alefragnani/vscode-project-manager/blob/v11.3.0/LICENSE.md)

One caveat is that it has an `"extensionDependency"` on a dependency that is not available on open-vsx:
```
	"extensionDependencies": [
		"ms-dotnettools.csharp"
	],
```
https://github.com/ionide/ionide-vscode-fsharp/blob/200e9484fc6ebf2dfe777ee276021517b7929c4e/release/package.json#L1568

A possible 'solution' is to remap `ms-dotnettools.csharp` to the open-vsx equivalent - `muhammad-sammy.csharp`, but perhaps you've encountered this before and have a more robust solution. Of course, the ideal would be to have the owner publish to open-vsx.

Related to https://github.com/onivim/oni2/issues/2974

Thank you!

### DIFF
--- a/extensions.json
+++ b/extensions.json
@@ -633,6 +633,10 @@
       "repository": "https://github.com/iocave/monkey-patch"
     },
     {
+      "id": "Ionide.Ionide-fsharp",
+      "repository": "https://github.com/ionide/ionide-vscode-fsharp"
+    },
+    {
       "id": "ipedrazas.kubernetes-snippets",
       "repository": "https://github.com/ipedrazas/kubernetes-snippets"
     },


spoenemann:
`ms-dotnettools.csharp` has a proprietary runtime license that allows usage only to Microsoft tools, so we shouldn't include it in open-vsx.org, which is made for non-Microsoft tools. Redirecting the dependency to an alternative might be an option, but we should check
 * that it works technically, and
 * that the alternative is really a proper replacement. We should be careful here, as in many tools the dependencies are installed automatically, and we should avoid surprising users by installing unwanted extensions.

Another possible option: Can the creators of `Ionide.Ionide-Fsharp` drop the dependency?

jankeromnes:
> A possible 'solution' is to remap `ms-dotnettools.csharp` to the open-vsx equivalent - `muhammad-sammy.csharp`, but perhaps you've encountered this before and have a more robust solution. Of course, the ideal would be to have the owner publish to open-vsx.

@bryphe Actually, I think you're right!

This is super-hacky and will probably break, but we could hot-patch the extension's `package.json` dependencies using a custom `"prepublish"` step, e.g. like so:

```json
    {
      "id": "Ionide.Ionide-fsharp",
      "repository": "https://github.com/ionide/ionide-vscode-fsharp",
      "location": "release",
      "prepublish": "sed -ri \"s_ms-dotnettools\.csharp_muhammad-sammy.csharp_g\" release/package.json"
    },
```

If this works, we can potentially re-use this trick for other 100% open-source extensions that unfortunately depend on other closed-sourced extensions that have an open-source equivalent.

jankeromnes:
Interesting development: https://github.com/muhammadsammy/ionide-vscode-fsharp/commit/186ab6279ddfb242340476001969aa6d2705f12b

bryphe:
Thanks @jankeromnes for the update!

I like your hot-patch idea - that would work in the meantime. Luckily it seems the dependency is more like an optional dependency, purely for debugger support (as far as I know,  unfortunately, the extension metadata doesn't have a way to model optional dependencies).

...but having a patched extension for F#, like we do for C#, is even better!

I'll close this out in anticipation of the patched F# extension being published here.

jankeromnes:
Aha! I was about to implement the nasty hack, but https://open-vsx.org/extension/Ionide/Ionide-fsharp was actually published to OpenVSX 6 days ago. 🎉

PR #278, opened 2021-02-01
Upgrades

jankeromnes:
Upgrades made with https://github.com/open-vsx/publish-extensions/pull/279 applied 💪


jankeromnes:
Allowed by https://github.com/open-vsx/publish-extensions/pull/279

PR #269, opened 2021-01-19
Add Jupyter extension

peat-psuwit:
This extension is a dependency for recent versions of ms-python.python.
Adding this should fix the publishing error for the Python extension.

The prepublish step is taken from the extension's CONTRIBUTE.md file.
PIP_USER=no is added to prevent its presence otherwise to intefere with
-t flag.

ms-toolsai.jupyter is available under MIT license.

PR #293, opened 2021-02-16
Build ms-python.python from source to avoid depending on Jupyter

peat-psuwit:
While we wait for Microsoft to clear up [if Open-VSX can distribute rebuilt-from-source Jupyter extension](https://github.com/microsoft/vscode-jupyter/issues/4814), it turns out that if we compile Python extension from source, we can exclude a step that adds the Jupyter extension as a dependency. So, this PR do just that.

As the steps to build the extension for distribution is quite long (it involves installing a few Python dependencies), I decided to factor them out to a dedicated script and refers to it in the `prePublish step`. As this requires referring back to this repo, an ability to do so is added to `publish-extension.js`.

I've tested the extension built using the script myself on [Coder's Code-server](https://github.com/cdr/code-server). The extension seems to still provide expected functionality, with (of course) Jupyter notebook support removed. This should allow Open-VSX users to have an up-to-date Python extension again.

Related: #269

### DIFF
--- a/extensions.json
+++ b/extensions.json
@@ -1036,9 +1036,9 @@
     {
       "id": "ms-python.python",
       "repository": "https://github.com/microsoft/vscode-python",
-      "version": "2021.1.502429796+nojupyter",
-      "checkout": "2021.1.502429796",
-      "prepublish": "%SELF%/prepublish-scripts/ms-python.python.sh 502429796",
+      "version": "2021.2.576481509+nojupyter",
+      "checkout": "2021.2.576481509",
+      "prepublish": "%SELF%/prepublish-scripts/ms-python.python.sh 576481509",
       "extensionFile": "ms-python-insiders.vsix"
     },
     {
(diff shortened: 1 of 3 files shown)


peat-psuwit:
Hmm, seems like the `npm install` step of this package requires `libpango1.0-dev` package. Not sure if GitHub Actions has the permission to install packages.

svenefftinge:
We should publish a full version of the python extension including the jupyter support.
See https://github.com/open-vsx/publish-extensions/pull/269#issuecomment-794108680

scollovati:
This is probably no more needed since the extension is optional (https://github.com/microsoft/vscode-python/pull/16267) and now has MIT license https://marketplace.visualstudio.com/items/ms-toolsai.jupyter/license

akosyakov:
I'm closing since both extensions are MIT now, and jupyter is optional dep, we just need to publish latest versions properly. Working on it right now.

PR #296, opened 2021-02-18
Add color theme "liqube.theme-liqube-dark".

houhoulis:
Thank you.

### DIFF
--- a/extensions.json
+++ b/extensions.json
@@ -810,6 +810,12 @@
       "repository": "https://github.com/LintangWisesa/VScode-Arduino-Theme"
     },
     {
+      "id": "liqube.theme-liqube-dark",
+      "repository": "https://github.com/liqube/liqube-dark-vscode",
+      "version": "1.0.5",
+      "checkout": "master"
+    },
+    {
       "id": "littlefoxteam.vscode-python-test-adapter",
       "download": "https://github.com/kondratyev-nv/vscode-python-test-adapter/releases/download/v0.6.6/vscode-python-test-adapter-0.6.6.vsix",
       "version": "0.6.6"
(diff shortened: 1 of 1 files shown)


extensions.json
```diff
@@ -809,6 +809,12 @@
       "id": "lintangwisesa.arduino",
       "repository": "https://github.com/LintangWisesa/VScode-Arduino-Theme"
     },
+    {
+      "id": "liqube.theme-liqube-dark",
+      "repository": "https://github.com/liqube/liqube-dark-vscode",
+      "version": "1.0.5",
+      "checkout": "master"
```
jankeromnes, line 816: Unfortunately, this repository doesn't provide any GitHub Releases or release tags. So, the best remaining option is to publish straight from the default branch (no need to explicitly check it out, and no need to pin a specific version):

```suggestion
      "repository": "https://github.com/liqube/liqube-dark-vscode"
```

jankeromnes (approved):
Approved, thanks!

Minor edit suggestion in-line, but I can apply it myself on GitHub before merging, so this is good to go. 🚀

houhoulis:
Thanks for the fix; sorry I missed it. 👍🏼

houhoulis:
I should mention, FYI, I generated this patch by signing up for Gitpod and using your Gitpod button tool. I'm not sure if that indicates a problem with the tool.

jankeromnes:
Sure! And no worries. Your usage of the tool was perfect, it's just a false positive (a rare case where the repository does not provide `.vsix` in GitHub Releases or suitable release tags -- but you're right, we could/should probably handle that corner case in the `add-extension` script).

jankeromnes:
Thanks again @houhoulis!

The extension was auto-published successfully: https://open-vsx.org/extension/liqube/theme-liqube-dark 🎉

Could you please confirm that it works as expected?

houhoulis:
It does! Much appreciated.

PR #297, opened 2021-02-18
Fix namespace of Symfony-code-snippets extension.

olifre:
It is nadim-vscode instead of nalabdou. 
Fixes the bad namespace introduced in c61f9ccad7dac6da445ad404384c5d586bd0650e .

### DIFF
--- a/extensions.json
+++ b/extensions.json
@@ -1107,7 +1107,7 @@
       "repository": "https://github.com/vshaxe/vshaxe.git"
     },
     {
-      "id": "nalabdou.Symfony-code-snippets",
+      "id": "nadim-vscode.Symfony-code-snippets",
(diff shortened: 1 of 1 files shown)


jankeromnes (approved):
Aha indeed, thanks for catching this @olifre!

In the future, please consider using the built-in helper script:

```
node add-extension https://github.com/nalabdou/Symfony-code-snippets
```

(This will automatically do the right thing.)

olifre:
> In the future, please consider using the built-in helper script:

Thanks, indeed, I checked this out now — very helpful :-). Will make use of that in future contributions (sadly, some other plugins I'd like to have have "forgotten" to license their code, so I'm now waiting on the developers to reply).

PR #298, opened 2021-02-18
vscode-makefile-tools added

GitMensch:
`node add-extension` created the download entry which is documented as "should be used if building is not feasible":

```json
"download": "https://github.com/microsoft/vscode-makefile-tools/releases/download/v0.2.0/makefile-tools.vsix"
```

but I think the plain build is better so used that instead.

### DIFF
--- a/extensions.json
+++ b/extensions.json
@@ -1061,6 +1061,11 @@
       "extensionFile": "dist/js-debug.vsix"
     },
     {
+      "id": "ms-vscode.makefile-tools",
+      "download": "https://github.com/microsoft/vscode-makefile-tools/releases/download/v0.2.0/makefile-tools.vsix",
+      "version": "0.2.0"
+    },
+    {
       "id": "ms-vscode.node-debug",
       "repository": "https://github.com/microsoft/vscode-node-debug",
       "version": "1.44.8",


jankeromnes:
Thanks @GitMensch!

> `node add-extension` created the download entry which is documented as "should be used if building is not feasible"

Hmm, where did you get this impression? Download is actually much preferred over building, because re-publishing an official `.vsix` carries much less risk of uploading a broken/non-working extension than attempting to repackage & rebuild an extension ourselves.

GitMensch:
> Hmm, where did you get this impression? 

That's already too old (fast pace here...) to be sure, I'll look out for this and if I see something report it as issue.

I _may_ have mixed this up with reproducible-issues: a download visx can contain anything and does not necessarily need to have (only) the code from the repo, there are a bunch of MS non-free-extensions that do this. Building from source means we can sure to publish/install what the code says...

But that's not my decision to take for this repo - if download is preferable then please feel free to edit and merge (I won't find the time to push a commit in the next hours).

GitMensch:
Changed (back) to the download variant.

> Hmm, where did you get this impression?

@jankeromnes Very likely from the README which gives the repo build version first and only after that says

> In cases where it is not feasible to build the extension from source, a download URL can be given instead

Which sounds to me like "that should only be done if really necessary".

jankeromnes:
> Changed (back) to the download variant in the hope to allow the merge of this one month old PR.

Many thanks! 🙏 Will approve and merge shortly.

> Very likely from the README as it shows the repo build version first and only after that and additional options "finally" says
> 
> > In cases where it is not feasible to build the extension from source, a download URL can be given instead
> 
> Which sounds to me like "that should only be done if really necessary".

Oh. That's an excellent point, I missed that part. I'll fix the README.

jankeromnes (approved):
Awesome, many thanks! 👍

jankeromnes:
Update: Auto-publishing worked! 🎉 https://open-vsx.org/extension/ms-vscode/makefile-tools

Thanks again for your contributions. 🙏

PR #303, opened 2021-02-21
Add "mindaro-dev.file-downloader"

paked:


extensions.json
```diff
@@ -898,6 +898,12 @@
       "repository": "https://github.com/mikestead/vscode-dotenv",
       "version": "1.0.1"
     },
+    {
+      "id": "mindaro-dev.file-downloader",
+      "repository": "https://github.com/microsoft/vscode-file-downloader",
+      "version": "1.0.10",
+      "checkout": "master"
```
jankeromnes, line 905: This repository has no GitHub Releases or dedicated release tag, so there is no need to try and pin or checkout a specific version.

The best we can do here is publish straight from the default branch.

```suggestion
      "repository": "https://github.com/microsoft/vscode-file-downloader"
```

jankeromnes (approved):
Thanks @paked! 👍 

One minor adjustment and this is good to go. 🚀

jankeromnes:
Update: Auto-publishing worked! 🎉 https://open-vsx.org/extension/mindaro-dev/file-downloader

@paked Please kindly confirm that the extension is working as intended.

PR #330, opened 2021-04-12
add Tyriar.sort-lines extension

ikim23:
I created this PR because creator of this extension [doesn't want](https://github.com/Tyriar/vscode-sort-lines/issues/92) to manage publishing to OVSX himself.

### DIFF
--- a/extensions.json
+++ b/extensions.json
@@ -1482,6 +1482,13 @@
       "version": "0.2.0"
     },
     {
+      "id": "Tyriar.sort-lines",
+      "repository": "https://github.com/Tyriar/vscode-sort-lines",
+      "version": "1.9.0",
+      "prepublish": "npx vsce package",
+      "extensionFile": "sort-lines-1.9.0.vsix"
+    },
+    {
       "id": "twxs.cmake",
       "repository": "https://github.com/twxs/vs.language.cmake"
     },
--- a/extensions.json
+++ b/extensions.json
@@ -1484,9 +1484,7 @@
     {
       "id": "Tyriar.sort-lines",
       "repository": "https://github.com/Tyriar/vscode-sort-lines",
-      "version": "1.9.0",
-      "prepublish": "npx vsce package",
-      "extensionFile": "sort-lines-1.9.0.vsix"
+      "prepublish": "yarn compile"
     },
     {
       "id": "twxs.cmake",


extensions.json
```diff
@@ -1481,6 +1481,13 @@
       "download": "https://github.com/tuxtina/json2yaml/releases/download/v0.2.0/json2yaml-0.2.0.vsix",
       "version": "0.2.0"
     },
+    {
+      "id": "Tyriar.sort-lines",
+      "repository": "https://github.com/Tyriar/vscode-sort-lines",
+      "version": "1.9.0",
```
jankeromnes, line 1487: Nit: When pinning a specific `"version"`, please also pin a specific `"checkout"` (e.g. release branch, tag, or commit ID), otherwise the script will publish from the default branch. (Alternatively, you can simply remove the `"version"` pin.)

extensions.json
```diff
@@ -1481,6 +1481,13 @@
...
+      "prepublish": "npx vsce package",
+      "extensionFile": "sort-lines-1.9.0.vsix"
```
jankeromnes, line 1489: Interesting, why are you calling `npx vsce package` here? (It is my understanding that `ovsx publish` will already do that.)

jankeromnes (commented):
Many thanks! Looks good, but I added two questions in-line. Please take a look when you have time.

extensions.json
ikim23: I removed `"version"`

extensions.json
ikim23: You are right. I got it wrong. Sorry.

According to `.github/workflows/release.yml` I changed `"prepublish"` to `"yarn compile"`. `"version"` and `"extensionFile"` fields are not needed, so I removed them.

jankeromnes: Awesome, many thanks for double-checking! 💯

extensions.json
```diff
@@ -1481,6 +1481,11 @@
       "download": "https://github.com/tuxtina/json2yaml/releases/download/v0.2.0/json2yaml-0.2.0.vsix",
       "version": "0.2.0"
     },
+    {
+      "id": "Tyriar.sort-lines",
+      "repository": "https://github.com/Tyriar/vscode-sort-lines",
+      "prepublish": "yarn compile"
```
jankeromnes, line 1487: Interesting. According to `vscode:prepublish` in the repository's [package.json](https://github.com/Tyriar/vscode-sort-lines/blob/c22a998b87c8ecdea84db5c8afe7bb8437e1f701/package.json#L11), `npm run compile` will also be run by `vsce` just prior to publishing.

However, since the repository's own [.github/workflows/release.yml](https://github.com/Tyriar/vscode-sort-lines/blob/c22a998b87c8ecdea84db5c8afe7bb8437e1f701/.github/workflows/release.yml#L25-L35) is indeed also set up to run both `yarn compile` followed by `npm run compile`, let's follow the same process here to avoid any subtle differences.

jankeromnes (approved):
Looks good to me, many thanks! 🎯

Still, it's kind of a shame that the author didn't want to fix https://github.com/Tyriar/vscode-sort-lines/issues/92, because the only two steps required would have been to:
1. Append the single line `ovsx publish -p $OVSX_TOKEN || true` at the bottom of [.github/workflows/release.yml](https://github.com/Tyriar/vscode-sort-lines/blob/c22a998b87c8ecdea84db5c8afe7bb8437e1f701/.github/workflows/release.yml)
2. Set up a token from [here](https://open-vsx.org/user-settings/tokens)

PR #335, opened 2021-04-22
Org preview

KlavsKlavsen:
adding org-preview module - now using node command :)

### DIFF
--- a/extensions.json
+++ b/extensions.json
@@ -1567,6 +1567,10 @@
       "checkout": "v11.4.0"
     },
     {
+      "id": "tootone.org-mode",
+      "repository": "https://github.com/vscode-org-mode/vscode-org-mode"
+    },
+    {
       "id": "vscode.cpp",
       "download": "https://openvsxorg.blob.core.windows.net/resources/vscode/cpp/1.44.2/vscode.cpp-1.44.2.vsix"
     },
--- a/extensions.json
+++ b/extensions.json
@@ -1567,8 +1567,10 @@
       "checkout": "v11.4.0"
     },
     {
-      "id": "tootone.org-mode",
-      "repository": "https://github.com/vscode-org-mode/vscode-org-mode"
+      "id": "vscode-org-mode.org-mode",
+      "repository": "https://github.com/vscode-org-mode/vscode-org-mode",
+      "version": "1.0.0",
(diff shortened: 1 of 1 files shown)


extensions.json
```diff
@@ -587,6 +587,12 @@
       "download": "https://github.com/golang/vscode-go/releases/download/v0.24.1/go-0.24.1.vsix",
       "version": "0.24.1"
     },
+    {
+      "id": "goodbaikin.org-preview",
+      "repository": "https://github.com/goodbaikin/VSC-Org-Preview",
+      "version": "1.0.0",
+      "checkout": "5245d553b65a97a86f5827968918e6a55444cc7d"
```
jankeromnes, line 594: Oops, apologies, in this case it's better to run just:

```
node add-extension https://github.com/goodbaikin/VSC-Org-Preview
```

Explanation: Since the repository was so kind to provide an official `.vsix` GitHub Release, it's much safer to re-publish that directly than to try and repackage the extension ourselves (and risk missing a step, leading to a broken extension package on OpenVSX).

Without a specific `--checkout` argument, `node add-extension <repo>` will check for available GitHub Releases, and automatically use them when available.

KlavsKlavsen: done - rewrote history this time :)

jankeromnes: Many thanks!

Huh, the script didn't do what I expected it to do. 🙄 Will look for bugs in it.

jankeromnes: Ah, found the culprit: The script didn't pick `'/releases/download/1.0.0/org-preview-1.0.0.vsix'` [because of the `-pre` in it](https://github.com/open-vsx/publish-extensions/blob/ebb7721d9d1dd8bf1182e8ccbe83de5a178b2c42/lib/github-scraper.js#L10) (thinking it's a pre-release 😅).

jankeromnes: @KlavsKlavsen I've fixed this bug in https://github.com/open-vsx/publish-extensions/commit/645a1ed5abe0be20f1f789f1ee28e024b87941d5 -- could you please rebase and re-run the script? 🙏

KlavsKlavsen: merged latest master and re-run :)

jankeromnes: Ah... 😁 perfection! Many thanks 🙏

jankeromnes (approved):
Looks perfect! 🎯 Many thanks for persisting. 🙏

PR #348, opened 2021-05-16
Add extensions for IBM i

chrjorgensen:


extensions.json
```diff
@@ -603,6 +607,11 @@
       "version": "1.2.0"
     },
     {
+      "id": "halcyontechltd.code-for-ibmi",
+      "repository": "https://github.com/halcyon-tech/code-for-ibmi",
+      "version": "0.4.2",
+      "checkout": "0.4.2"
```
jankeromnes, line 613: Simple curiosity, why pin `0.4.2` here instead of the latest available release `0.4.6`?

```suggestion
      "version": "0.4.6",
      "checkout": "0.4.6"
```

jankeromnes (approved):
Looks good to me, many thanks! 🎯

One small question in-line. 👇

extensions.json
chrjorgensen: 0.4.2 was the available version at the time when the PR was build. I know the developer made a few changes last night, so he probably bumped the version a few times since then.

Will the CI get the new version automatically? Or do I have to request any new versions after this?

jankeromnes: Thanks for clarifying @chrjorgensen!

I'm happy to go with any version you pick (please either click on "Commit suggestion" or simply resolve this discussion -- after that I'll merge this Pull Request).

There is an auto-upgrade job that attempts to look for more recent releases/tags on every Sunday (see for example https://github.com/open-vsx/publish-extensions/pull/346).

jankeromnes:
Thanks again! This is good to merge. 🚀

PR #352, opened 2021-05-30
Swellaby.vscode-rust-test-adapter: Add prepublish step to build js files

MJDSys:
This extension hasn't had a version bump in a while, and it's not clear when the next version will be released.  Is it possible to update the contents of this extension on Open VSX with this change?  Otherwise the extension as is does not function (it's missing all of its javascript application code).

### DIFF
--- a/extensions.json
+++ b/extensions.json
@@ -1476,7 +1476,8 @@
     },
     {
       "id": "Swellaby.vscode-rust-test-adapter",
-      "repository": "https://github.com/swellaby/vscode-rust-test-adapter"
+      "repository": "https://github.com/swellaby/vscode-rust-test-adapter",
+      "prepublish": "npm update vscode && npm install && npm run build"
     },
     {
       "id": "tchayen.markdown-links",


extensions.json
```diff
@@ -1476,7 +1476,8 @@
     },
     {
       "id": "Swellaby.vscode-rust-test-adapter",
-      "repository": "https://github.com/swellaby/vscode-rust-test-adapter"
+      "repository": "https://github.com/swellaby/vscode-rust-test-adapter",
+      "prepublish": "npm update vscode && npm install && npm run build"
```
jankeromnes, line 1480: Ah, unfortunately, I don't think a `prepublish` step will fix the failing `npm install`, because events actually happen in this order:

1. recursively clone the extension repository
2. if there is a `yarn.lock` file, run `yarn install`, otherwise run `npm install`
3. next run any `prepublish` step

See also:

https://github.com/open-vsx/publish-extensions/blob/eee429cb14e55f23132a1e6c8192da6ce876271e/publish-extensions.js#L113-L124

jankeromnes (approved):
Thanks again! I'm happy to merge this `prepublish` step, because auto-publishing this extension is currently broken, and any attempt to fix it is helpful.

However, if `npm install` is indeed currently broken, I don't think this approach will work (see below for details).

jankeromnes:
Ah, the `npm install` error is actually the same as https://github.com/open-vsx/publish-extensions/pull/375#issuecomment-868992787

Would be cool to find a proper solution to this.

calebcartwright:
Maintainer of the aforementioned extension here, and just a heads up that i've opened a request to claim our namespace.

I'm not really sure what all is going on with this packaging process, but hope we'll have access to publish our own extension soon. We've never had any issues installing deps, generating the vsix, and uploading it places, so I'm optimistic it'll be easy for us to start uploading our vsix on this registry too.

brianking:
@calebcartwright Great, once you are ready here is a handy reference for publishing:
https://github.com/eclipse/openvsx/wiki/Publishing-Extensions

calebcartwright:
Thanks @loujaybee. As discussed in https://github.com/EclipseFdn/open-vsx.org/issues/649 I've decided against trying to directly publish because I'm not comfortable with the requirements and access grants mandatory to make that happen (realize that may elicit some eyerolls but just wanted to be transparent that we're not going that route :smile:). Instead, I'd like to see what will be necessary to get your publish process working with our extension.

It looks like the root cause of the build/package failures in the existing process were driven by the fact that we hadn't updated the extension to account for the VS Code team's switch from/splitting of `vscode` into the respective new dependencies triggering the npm failure.

I've addressed that, so I'm cautiously optimistic that will resolve the process here or at least help move things forward.

I will echo my comment from the thread in our upstream repo that engaging directly with maintainers is a good first step, and I think it'd be helpful to include that in documentation/guidance on getting new extensions incorporated into the Open VSX ecosystem. Additionally, I also think it'd be useful to provide some documentation that describes your publishing process so that maintainers, like myself, can edify ourselves and have a better understanding of what's being done/what may be going wrong with our respective extensions.

For example, I've been moving to Node 16/npm 7 in the majority of my projects and was planning on doing the same with the Rust test adapter extension discussed on this thread. However, I realized after inspecting your workflow logs that doing so would be problematic given the different lockfile formats between npm 6 that's bundled with node 14 and utilized in your publishing process.

IMO something that describes how extensions are being taken from source and packaged/bundled, and the characteristics of the environment where that is being performed, would be extremely helpful.

akosyakov:
@calebcartwright the easies if you produce a GitHub release with vsix file when we can directly consume it, see for instance: https://github.com/open-vsx/publish-extensions/blob/32c858eccb978d48dc0e507fd04c4825d240d3d0/extensions.json#L216-L220 We have an upgrade script then which monitors GitHub releases for you repo and auto upgrades to new.

calebcartwright:
https://github.com/open-vsx/publish-extensions/pull/478

PR #356, opened 2021-06-10
Add felipecaputo.git-project-manager

1efty:
Here is the issue asking the maintainer: https://github.com/felipecaputo/git-project-manager/issues/115

Please consider adding this.

### DIFF
--- a/extensions.json
+++ b/extensions.json
@@ -510,6 +510,11 @@
       "checkout": "v1.1.2"
     },
     {
+      "id": "felipecaputo.git-project-manager",
+      "download": "https://github.com/felipecaputo/git-project-manager/releases/download/v1.5.0/git-project-manager-1.5.0.vsix",
+      "version": "1.5.0"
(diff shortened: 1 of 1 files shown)


jankeromnes (approved):
Thanks! While waiting for the maintainer's response, I'm happy to temporarily publish this extension from here.

One small thing though: Why is there a `checkout` parameter? Normally when you provide a VSIX direct `download` URL, there is a `version` but no `checkout`.

Did you add this extension by running `node add-extension https://github.com/felipecaputo/git-project-manager`? (Normally this will auto-detect everything and update `extensions.json` with a correct configuration -- but if it adds a `checkout` here, that's a bug.)

1efty:
@jankeromnes I added that manually. I've amended the commit using the method you outlined. Appreciate the prompt response.

jankeromnes (approved):
Very cool, thanks for fixing!

PR #388, opened 2021-07-08
Remove the git-client-ibmi extension

mattburgess:
The git-client-ibmi extension has a misnamed license file which
causes the upgrade-extensions workflow to fail (see
https://github.com/open-vsx/publish-extensions/runs/2981924283?check_suite_focus=true).

An upstream PR has been raised to correct the filename (see
https://github.com/halcyon-tech/git-client-ibmi/pull/1) but
until that has been merged then upgrade-extensions will never
succeed. So, temporarily remove the extension.

### DIFF
--- a/extensions.json
+++ b/extensions.json
@@ -649,12 +649,6 @@
       "checkout": "0.5.8"
     },
     {
-      "id": "halcyontechltd.git-client-ibmi",
-      "repository": "https://github.com/halcyon-tech/git-client-ibmi",
-      "version": "0.0.4",
-      "checkout": "0.0.4"
-    },
-    {
       "id": "HansUXdev.bootstrap5-snippets",
       "repository": "https://github.com/HansUXdev/B5-Snippets",
       "prepublish": "npm run concat"


chrjorgensen:
I was the one adding the git-client-ibmi extension to Open-VSX, so I'm partly responsible. I also added the vscode-ibmi extension by the same author and stumbled across the same problem with a "misnamed license file", and even made a PR to have it spelled "correctly".

Seeing the answer from the author on the upstream PR made me curious, and although English is not my main language, googling the word "licence" shows, that this is the British spelling of the word when used as a Noun, where the American spelling is "license".

IMHO the solution is not to have all British VS Code extensions rename their LICENCE file to LICENSE - especially when it seems like GitHub is accepting the British spelling and showing the license tag under "About". The solution ought to be to have Open-VSX publish-extension script to accept both spellings - like GitHub does.

I have no knowledge of GitHub workflows, but looking at the latest run it seems to me like the git-client-ibmi extension is not the cause of the failure: The process continues after the extension and tries the next one and so on. So are you sure this extension really is a problem for the publishing workflow?

jankeromnes:
> IMHO the solution is not to have all British VS Code extensions rename their LICENCE file to LICENSE - especially when it seems like GitHub is accepting the British spelling and showing the license tag under "About". The solution ought to be to have Open-VSX publish-extension script to accept both spellings - like GitHub does.

Agreed! 😊 Fixing this in https://github.com/eclipse/openvsx/pull/307

> An upstream PR has been raised to correct the filename (see
> halcyon-tech/git-client-ibmi#1) but
> until that has been merged then upgrade-extensions will never
> succeed. So, temporarily remove the extension.

Many thanks & good catch! However, we can keep the extension listed, and simply exclude it from auto-upgrades temporarily by adding it here:

https://github.com/open-vsx/publish-extensions/blob/decfe1b8f7d0383b94e300694d8dc83e35f36c5d/upgrade-extensions.js#L19-L37

I'll do that now.

jankeromnes:
Update: Done in https://github.com/open-vsx/publish-extensions/pull/392 (and I've also manually run `node upgrade-extensions` in order to catch up)

Thanks again for identifying & reporting this problem! 💯

PR #394, opened 2021-07-21
Add Android Debugging Support extension.

benjaminwolkchen:


extensions.json
```diff
@@ -16,6 +16,10 @@
       "id": "adamhartford.vscode-base64",
       "repository": "https://github.com/adamhartford/vscode-base64"
     },
+    {
+      "id": "adelphes.android-dev-ext",
+      "repository": "https://github.com/adelphes/android-dev-ext"
```
jankeromnes, line 21: It's probably safer to publish a specific tag/version than to always publish from the default branch:

```suggestion
      "repository": "https://github.com/adelphes/android-dev-ext",
      "version": "1.3.2",
      "checkout": "v1.3.2"
```

jankeromnes (approved):
Many thanks! Looks good to me, with one suggestion (that I can apply in one click, so no extra work required 🌴)

PR #404, opened 2021-08-07
Extention: Added vscode-ungit

JRedOW:
Added Ungit integration in VSCode.  https://github.com/Hirse/vscode-ungit

### DIFF
--- a/extensions.json
+++ b/extensions.json
@@ -715,6 +715,7 @@
     {
       "id": "Hirse.vscode-ungit",
       "download": "https://github.com/Hirse/vscode-ungit/releases/download/v2.2.3/vscode-ungit-2.2.3.vsix",
+      "version": "2.2.3"
     },
     {
       "id": "hollowtree.vue-snippets",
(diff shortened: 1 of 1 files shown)


extensions.json
```diff
@@ -712,6 +712,10 @@
       "download": "https://github.com/lgalabru/clarity-lsp/releases/download/v0.5.1/clarity-lsp.vsix",
       "version": "0.5.1"
     },
+    {
+      "id": "Hirse.vscode-ungit",
+      "download": "https://github.com/Hirse/vscode-ungit/releases/download/v2.2.3/vscode-ungit-2.2.3.vsix",
```
jankeromnes, line 717: With `"download"` as with `"checkout"`, it's better to add an appropriate `"version"`, otherwise the script will always try to re-publish the same version every night. (It's an optimization.)

```suggestion
      "download": "https://github.com/Hirse/vscode-ungit/releases/download/v2.2.3/vscode-ungit-2.2.3.vsix",
      "version": "2.2.3"
```

jankeromnes (approved):
Many thanks for this great addition! 🎉 

Code change looks good to me, except for one minor detail, which I can probably adjust myself before merging. 🚀

PR #409, opened 2021-08-22
add extension: cmake-format

playgithub:


playgithub:
Prerequest: `pip install cmakelang`
So what if python has not been installed? e.g. no python by default on Windows

jankeromnes (approved):
Looks great :sparkles: many thanks!

> Prerequest: `pip install cmakelang`
> So what if python has not been installed? e.g. no python by default on Windows

Python isn't required to re-publish an official `.vsix` extension file to OpenVSX.

And, for users of this extension, I believe the situation will be the same whether they're using the Microsoft marketplace or OpenVSX (I would expect the extension to try and detect Python, and show an error if it's not available -- but this is out of scope here, and could be raised with the extension's developers instead).

PR #413, opened 2021-08-23
remove explicit version for ms-vscode.hexeditor

GitMensch:
... as this is very outdated (the extension does not do release-tagging).

In the hope to get it updated from 1.3.0 to 1.8.2.


GitMensch:
OK, that worked, the extension's version 1.8.2 is now also available.

PR #425, opened 2021-09-02
Publish vscode-smoketest-check as web

jeanp413:


jankeromnes (approved):
Should work, thanks!

FYI, this will only apply to new versions, unless we delete the current latest version from https://open-vsx.org

PR #428, opened 2021-09-02
set timeout 30mins for ms-toolsai.jupyter

akosyakov:
it takes a while to webpack everything


akosyakov:
Let's wait merging till I manually publish proper versions first.

akosyakov:
I published proper version, it is good to merge now.

PR #435, opened 2021-09-13
Add goessner.mdmath extension

Tim-Tech-Dev:


Tim-Tech-Dev:
There is already an  [issue](https://github.com/goessner/mdmath/issues/95) in authors repository. Someone more qualified could look into helping him publish it on open-vsx. Thanks

PR #436, opened 2021-09-13
add extension volar

OnlyBjarne:


extensions.json
```diff
@@ -863,6 +863,10 @@
       "id": "johnpapa.winteriscoming",
       "repository": "https://github.com/johnpapa/vscode-winteriscoming"
     },
+    {
+      "id": "johnsoncodehk.volar",
+      "repository": "https://github.com/johnsoncodehk/volar"
```
jankeromnes, line 868:
```suggestion
      "repository": "https://github.com/johnsoncodehk/volar",
      "version": "0.27.23",
      "checkout": "v0.27.23",
      "prepublish": "yarn compile",
      "location": "extensions/vscode-vue-language-features"
    },
    {
      "id": "johnsoncodehk.vscode-typescript-vue-plugin",
      "repository": "https://github.com/johnsoncodehk/volar",
      "version": "0.27.23",
      "checkout": "v0.27.23",
      "prepublish": "yarn compile",
      "location": "extensions/vscode-typescript-vue-plugin"
```

jankeromnes (approved):
Many thanks for this cool addition @OnlyBjarne!

Looking at the repository, I think the extensions are actually inside the [extensions](https://github.com/johnsoncodehk/volar/tree/v0.27.23/extensions) directory, which needs to be reflected in a `"location"` attribute here. Also, publishing a specific `"version"` and tag (`"checkout"`) is recommended when possible.

Finally, I'm not 100% sure about this one, but looking at the repository's general structure and its [scripts](https://github.com/johnsoncodehk/volar/blob/v0.27.23/package.json#L7), it seems safer to me to add a `"prepublish": "yarn compile"` to build the entire repository prior to publishing the extensions.

I'll make both adjustments as suggestions in-line, which I can auto-apply before merging, so no extra work required here. 🙌 🌴

yumcc-com:
Hello, Is the work of publishing open-vsx still going on?

akosyakov:
Please send a new PR. 

> Thanks for your response! I prefer to hand over the release of OpenVSX to this repo, and I am also willing to maintain the Volar release settings of this repo, so in my repo I can focus on solving issues instead of dealing with additional CI and token management.

@johnsoncodehk It would be ideal that we don't need this repo anymore. Initially it was used to seed the registry, but now there are many authors doing publishing directly while the registry adopted by all forks of VS Code, i.e. Gitpod, Theia, VSCodium, code-server and so on. Just from Gitpod is used by 400 thousands of devs. Technically it should be straightforward. You have to get a token and call ovsx publish with it, for instance how RedHat is publishing the java extension: https://github.com/redhat-developer/vscode-java/blob/ff51dbffc30dd23190335c929e9f34ccdf511554/Jenkinsfile#L73-L82 Docs are here: https://github.com/eclipse/openvsx/wiki/Publishing-Extensions

PR #438, opened 2021-09-15
Add headwind extension

kalin:
Adding the headwind extension until the author has time to submit future releases directly.

See: https://github.com/heybourn/headwind/issues/95


jankeromnes (approved):
Thanks, looks good to me!

PR #445, opened 2021-09-23
Add tomorrow theme

gbraad:
Not sure if this is correct, as the themes are all collected in this repo but divided into subfolders.

https://marketplace.visualstudio.com/items?itemName=ms-vscode.Theme-TomorrowKit

License: MIT

### DIFF
--- a/extensions.json
+++ b/extensions.json
@@ -1979,6 +1979,11 @@
     {
       "id": "zxh404.vscode-proto3",
       "repository": "https://github.com/zxh0/vscode-proto3"
+    },
+    {
+      "id": "ms-vscode.Theme-TomorrowKit",
+      "repository": "https://github.com/microsoft/vscode-themes/",
+      "location": "tomorrow"
     }
   ]
 }


jankeromnes (approved):
Thanks @gbraad!

> Not sure if this is correct, as the themes are all collected in this repo but divided into subfolders.

The syntax you've used looks correct. 👍 Also, if you wanted to add more themes, you'd need to add other similar entries (with different `"id"` and `"location"` but the same `"repository"`).

gbraad:
>  you'd need to add other similar entries

I figured. Let's see if this works as expected. After this I will created the other entries.

jankeromnes:
> I figured. Let's see if this works as expected. After this I will created the other entries.

Works for me, thanks!

Results should come in tonight at 03:03 UTC, so we'll know tomorrow whether publishing worked.

jankeromnes:
Also, on a side note, you can add the other extensions like so:

```
node add-extension https://github.com/microsoft/vscode-themes --location=tomorrow
```

(This will automatically figure out the `"id"` and sort the list alphabetically 😇)

PR #454, opened 2021-10-05
Upgrade IBM i extensions to current version

chrjorgensen:


jankeromnes (approved):
Thanks for catching this!

Note: We've recently fixed our auto-upgrade script (was broken for a few weeks), and it already upgraded two of these three extensions: https://github.com/open-vsx/publish-extensions/commit/7dd5625b65059c2e91e08e76864f64ef8bc6b7c6

For `halcyontechltd.git-client-ibmi`, we've noticed that it's currently ignored by the the upgrade script, but that was because of a problem that is now fixed:

https://github.com/open-vsx/publish-extensions/blob/187bdb17f2dd7d11a66bb25b03d4c32723c33e14/upgrade-extensions.js#L36

So, we can also resume auto-upgrades for this extension 🙌

PR #458, opened 2021-10-12
fix: only report validation fails to Slack when on the default branch

jankeromnes:


.github/workflows/validate-extensions.yml
```diff
@@ -14,7 +14,7 @@ jobs:
     - run: npm install
     - run: node validate-extensions
     - name: Slack Notification
-      if: always()
+      if: github.ref == 'refs/heads/master'
```
jankeromnes, line 17: From @akosyakov [on Slack](https://gitpod.slack.com/archives/C01KGM9BH54/p1634028035201000?thread_ts=1634025992.198300&cid=C01KGM9BH54) (internal):

> always() though is important, it says that report even if previous steps failed, so it should not be removed

jankeromnes: Many thanks! I don't really understand these conditional workflow steps yet.

I've applied your suggestion ✅ 🚀

PR #461, opened 2021-10-16
Allow vscode-ibmi-languages to update instead of fixed version

chrjorgensen:


chrjorgensen:
@jankeromnes There is a problem with your SLACK notification setup, which makes my PR check fail:
```
[ERROR] Secret `SLACK_WEBHOOK` is missing. Please add it to this action for proper execution.
```

PR #472, opened 2021-10-27
Use the GitHub GraphQL API instead of the scraper

### LINKED ISSUE #471: Don't publish prereleases
First discussed in https://github.com/open-vsx/publish-extensions/pull/466#discussion_r734630003.

## Changes
The script scrapes the GitHub releases page and gets the release URL, instead we should fetch the GitHub Releases API (https://docs.github.com/en/rest/reference/repos#releases) and look for the `prerelease` key to determine whether to publish the version or not. 

### Things to look out for
- Some repositories maybe ship prerelease builds as preview versions.
- The GitHub API quota (https://docs.github.com/en/rest/reference/rate-limit)

filiptronicek:
Fixes #471

## How to test
Run `GH_PAT="GH_PERSONAL_ACCESS_TOKEN" node upgrade-extensions.js` (replace `GH_PERSONAL_ACCESS_TOKEN` with a [GitHub PAT](https://docs.github.com/en/authentication/keeping-your-account-and-data-secure/creating-a-personal-access-token), the PAT does not have to have any additional scopes added to it)


filiptronicek:
> @filiptronicek besides tag based issues, I found one regression for "Unity.unity-debug".
> 
> What update produces: Screenshot 2021-10-29 at 12 19 30
> 
> How releases look like: [Unity-Technologies/vscode-unity-debug/releases](https://github.com/Unity-Technologies/vscode-unity-debug/releases) What is published to MS marketplace: [marketplace.visualstudio.com/items?itemName=Unity.unity-debug](https://marketplace.visualstudio.com/items?itemName=Unity.unity-debug)
> 
> It seems that they don't do releases anymore, we should switch to use GitHub tags instead I guess: [Unity-Technologies/vscode-unity-debug@`Version-3.0.2` (release)](https://github.com/Unity-Technologies/vscode-unity-debug/releases/tag/Version-3.0.2)

It seems like they still do releases, @akosyakov, they just didn't mark the 3.x version as a full release, this doesn't look like a bug on our side 🤨: 
image

.github/workflows/upgrade-extensions.yml
```diff
@@ -23,6 +23,8 @@ jobs:
     - run: npm install
     - run: npm install --global yarn
     - run: node upgrade-extensions
+      env: 
+        GH_PAT: ${{ secrets.GITHUB_TOKEN }}
```
loujaybee, line 27: Could we update the secret name to remove reference to the personal access token?

lib/getReleases.js
```diff
@@ -0,0 +1,57 @@
...
+              }
+            }
+          }
+        }
+      }
+    }` }),
+  });
+
+  if (!releases.ok) {
+    if (releases.status === 401) {
```
loujaybee, line 40: Could we consolidate the nested if's with an `&&`?

lib/getReleases.js
```diff
@@ -0,0 +1,57 @@
...
+      throw new Error("The GitHub personal access token is invalid");
```
loujaybee, line 41: I like the custom errors, could we also log the error object here, also?

akosyakov:
> It seems like they still do releases, @akosyakov, they just didn't mark the 3.x version as a full release, this doesn't look like a bug on our side 🤨:

If we produce our script is producing extension.json with versions which are less than already published to OpenVSX or newer than to MS Marketplace it is our issue. We should here either switch from GitHub releases to tags or go and fix in the extension repo. I doubt the latter is easy.

.github/workflows/upgrade-extensions.yml
filiptronicek: Donezo, renamed it in [`ded9bb4` (#472)](https://github.com/open-vsx/publish-extensions/pull/472/commits/ded9bb4b0b487396e0cdc77bc134807ea9abac38)

akosyakov:
```
Fetching yzane/vscode-markdown-pdf to check for updates...
yzane.markdown-pdf: failed to upgrade downloads: Error: The latest release from GitHub (1.4.4) is older than the one already in the extensions list 1.4.4
    at Object.exports.findLatestVSIXRelease (/workspace/publish-extensions/lib/getReleases.js:61:11)
    at runMicrotasks (<anonymous>)
    at processTicksAndRejections (node:internal/process/task_queues:96:5)
    at async /workspace/publish-extensions/upgrade-extensions.js:90:24
```

It looks wrong right? It should be failed, but rather nothing happens.

filiptronicek:
@akosyakov, I made it so if the version matches it won't output an error, rather having to be outdated to fail. I have some extensions that are not setup correctly, I will create a PR to fix them.

.github/workflows/upgrade-extensions.yml
```diff
@@ -38,3 +40,6 @@ jobs:
       env:
         SLACK_WEBHOOK: ${{ secrets.GITPOD_SLACK_WEBHOOK }}
         SLACK_COLOR: ${{ job.status }}
+    - name: Report failed extensions
+      run: node report-extensions
+
```
bigint, line 45:
```suggestion
      run: node report-extensions
```

There was two EOL 😄

bigint (changes requested):

PR #478, opened 2021-11-04
use VSIX download for Rust Test Adapter

calebcartwright:
The existing package and publish process has been broken for a while, and should finally be resolved with this.

Refs https://github.com/open-vsx/publish-extensions/pull/352 and https://github.com/EclipseFdn/open-vsx.org/issues/649


corneliusludmann (approved):
Looks good. Thanks!

PR #483, opened 2021-11-08
docs: update README with reasons for publishing extensions at source

loujaybee:
This change updates the README to try to explain some of the reasons why publishing at a source repo is better than within `publish-extensions`. Any further reasoning, or wording updates are very welcome!

### DIFF
--- a/README.md
+++ b/README.md
@@ -7,7 +7,13 @@ A CI script for publishing open-source VS Code extensions to [open-vsx.org](http
 
 ## When to Add an Extension?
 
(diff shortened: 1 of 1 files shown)


loujaybee:
CC: @filiptronicek we should integrate some of these reasons into the "issue template" that we share with the repo owners to make it clear why it's important that they should set up the publishing, if possible 🙏 🙏 🙏

akosyakov (approved):

loujaybee:
Thanks @akosyakov !

PR #492, opened 2021-11-10
Major refactor of the publishing script

filiptronicek:
## Changes in this PR

- [x] Clone only the last commit of the repository
- [ ] Always try to use the extension's version from the MS Marketplace

### DIFF
--- a/lib/getReleases.js
+++ b/lib/getReleases.js
@@ -15,12 +15,49 @@ class GitHubAPIError extends Error {
   }
 }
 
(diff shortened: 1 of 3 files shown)


akosyakov:
I don’t expect that the publish job should update extensions.json file, but rather we should these file to remove such information and compute on the flight.

akosyakov:
@filiptronicek I thought we agreed to restrain one big refactoring PR, but refactor it in several PRs. First step to pass msversion to publish script and make use of it.

PR #494, opened 2021-11-11
For extension with a `.vsix` file, publish the MS marketplace version

filiptronicek:
For extensions that use a release asset as their source for publishing are going to be always published by the same version as is present on the Visual Studio Marketplace.

## Pre-merge checklist
- [x] Repositories that publish neither releases, nor tags anymore should be built from source
- [x] in the `lastCommit` case in `getReleases.js`, publish the commit-associated version instead of the Microsoft one.
- [x] if there's no `download`, but an `extensionFile`, don't build the extension
- [x] Try to delete `version`, see if that has any bigger implications / breaks any extensions

## How to test 
Run `GITHUB_TOKEN="GH_PERSONAL_ACCESS_TOKEN" SKIP_BUILD=1 node publish-extensions` (replace GH_PERSONAL_ACCESS_TOKEN with a GitHub PAT, the PAT does not have to have any additional scopes added to it) and inspect the logs.

fixes https://github.com/open-vsx/publish-extensions/issues/475


publish-extensions.js
```diff
@@ -135,8 +136,10 @@ const flags = [
       }
 
       let timeout;
-      await new Promise((resolve, reject) => {
-        const p = cp.spawn(process.execPath, ['publish-extension.js', JSON.stringify(extension)], {
+      await new Promise(async (resolve, reject) => {
+        const repository = extension.download && extension.download.replace(/\/releases\/download\/.*$/, '');
```
akosyakov, line 140: Let's fail if msVersion is defined, but extension.repository is undefined. It means that we should manually add repository in extensions.json.

lib/getReleases.js
```diff
@@ -15,14 +15,51 @@ class GitHubAPIError extends Error {
...
+    }
+
+    const releaseData = await release.json();
+
+    if (releaseData?.data.repository?.release) {
+    return releaseData?.data.repository?.release.releaseAssets.edges.
+      map((/** @type {{ node: {downloadUrl: any;} }} */ a) => a.node.downloadUrl).
+      find((/** @type {string} */ d) => d.match(/\/releases\/download\/[-._a-zA-Z0-9\/%]*\.vsix/g));
+    } else {
+      console.warn("Specified release not found, defaulting to latest release...");
```
akosyakov, line 59: if msVersion is specified we should make sure that we don't use anything else, so throw or return undefined here.

akosyakov:
How can one test it?

filiptronicek:
@akosyakov added how-to-test instructions, thanks for pointing that out!

publish-extensions.js
```diff
@@ -135,8 +136,19 @@ const flags = [
...
+        if (msVersion && (!repository && !extension.repository)) {
+          reject(new Error("Repository not found, you should add it to extensions.json"));
+        }
+
+        const vsixDownloadLink = extension.download && await getReleases.findLatestVSIXRelease(repository || extension.repository, extension.version, msVersion);
```
akosyakov, line 145: Do we really need download guard?

publish-extensions.js
```diff
@@ -135,8 +136,19 @@ const flags = [
       }
 
       let timeout;
-      await new Promise((resolve, reject) => {
-        const p = cp.spawn(process.execPath, ['publish-extension.js', JSON.stringify(extension)], {
+      await new Promise(async (resolve, reject) => {
+        const repository = extension.download && extension.download.replace(/\/releases\/download\/.*$/, '');
```
akosyakov, line 140: Please just remove it, write code which does not rely or download properly

publish-extensions.js
```diff
@@ -135,8 +136,19 @@ const flags = [
...
+        if (msVersion && (!repository && !extension.repository)) {
+          reject(new Error("Repository not found, you should add it to extensions.json"));
+        }
+
+        const vsixDownloadLink = extension.download && await getReleases.findLatestVSIXRelease(repository || extension.repository, extension.version, msVersion);
+        if(repository && !vsixDownloadLink) {
```
akosyakov, line 146: We should fail if repository is not provided before trying to check for release no?

akosyakov:
@filiptronicek I did not know whether it is fine to push to this PR and opened another PR against it: https://github.com/open-vsx/publish-extensions/pull/498

It properly addresses the feedback, then these changes are applied one has to run following command:
```
DRY_RUN=1 node publish-extensions
```

It will validate extensions.json for missing repository property and for cases when vsix cannot be resolved from releases. We should:
- update extension.json to add repository property and delete download
- look why vsix cannot be resolved and figure out how to handle it instead

filiptronicek:
Migration to the repository property should be working from my testing, but some repos neither create tags nor do they publish releases, they just bump their package.json version (e.g. https://github.com/alefragnani/vscode-bookmarks, https://github.com/alefragnani/vscode-project-manager). We should have a case for that also.

image

/cc @akosyakov

akosyakov:
@filiptronicek could you compile list of them please? 🙏

publish-extensions.js
```diff
@@ -134,9 +139,49 @@ const flags = [
         continue;
       }
 
+      if (msVersion) {
+        if (!extension.repository) {
+          if (extension.download) {
```
akosyakov, line 144: We should update extensions.json don’t do any magic in scripts.

filiptronicek:
> @filiptronicek could you compile list of them please? 🙏

@akosyakov here it is: https://filiptronicek.notion.site/Extensions-with-weird-publish-processes-03d28d2ebda54c08810856b6adf8ba4e

akosyakov:
I’m confused I don’t see changes to extensions.json file, but see code which using download property again. I don’t want any check or smartness on download property. It should not appear in extensions.json anymore.

publish-extensions.js
```diff
@@ -134,9 +137,42 @@ const flags = [
         continue;
       }
 
+      if (msVersion) {
+        if (!extension.repository) {
+          if (extension.download) ()
```
akosyakov, line 142: We care about all extensions not about only with download property.

publish-extensions.js
```diff
@@ -134,9 +137,42 @@ const flags = [
...
+      }
+
+      if (Boolean(process.env.DRY_RUN)) {
+        continue;
+      }
+
       let timeout;
-      await new Promise((resolve, reject) => {
-        const p = cp.spawn(process.execPath, ['publish-extension.js', JSON.stringify(extension)], {
+      await new Promise(async (resolve, reject) => {
```
akosyakov, line 158: Please revert changes to this promise. We should go manually through failed dry run and adjust extensions.json

lib/getReleases.js
```diff
@@ -1,26 +1,95 @@
...
+        }` }),
+    });
+    if (!response.ok) {
+      throw new Error(response.statusText);
+    }
+
+    const result = await response.json();
+    if (!result.data && result.errors) {
+      throw new Error(JSON.stringify(result.errors));
+    }
```
bigint, line 50: **suggestion (non-blocking):** Why not use `try catch` instead of checking errors in the if condition?

publish-extensions.js
filiptronicek: Sorry for that one, this one was left there for some reason, but a mistake for sure.

lib/getReleases.js
filiptronicek: Well if we wanted to try catch, we would need to raise the exception somewhere, conditionally, this cannot be done in a `try catch` block.

akosyakov:
`DRY_RUN=1 node publish-extensions` still does not succeed, we need to make it green before merging

extensions.json
```diff
@@ -54,13 +54,15 @@
     },
     {
       "id": "alefragnani.Bookmarks",
-      "download": "https://github.com/alefragnani/vscode-bookmarks/releases/download/v13.0.1/Bookmarks-13.0.1.vsix",
-      "version": "13.0.1"
+      "version": "13.2.2",
```
akosyakov, line 57: I'm not sure about specifying versions, see internal discussions: https://gitpod.slack.com/archives/CQJLAF3SR/p1637060685005600

I think we should rather read version from package.json of cloned git repo or downloaded vsix file.

akosyakov: Hm, turned out that some extensions forget to push updated version to GitHub, but only to Marketplace, so in such cases we should provide commit and proper version.

GitMensch:
That is an interesting approach and puts more trust to the published vsix, trusting that the visx actually matches the source.
It definitely cuts down necessary resources and most users likely trust the extension writers in any case; but I _still_ think that it would be good to announce change like that.

Questions:
1. "publish the MS marketplace version" - it actually doesn't but instead publishes the vsix found under the releases, correct?
2. Does this only apply to GitHub hosted repos?
3. Doesn't the remove of `checkout` likely break a bunch of extensions that have master/main as "dev snapshot" and a stable branch which was previously explicit configured via `checkout`? Is there a new way of doing the same?

akosyakov:
> That is an interesting approach and puts more trust to the published vsix, trusting that the visx actually matches the source.

Before we just published whatever was on main/master and used checkout to resolve conflicts. In fact some code which is not published to MS but was experimental ended up on OpenVSX. With new approach we try to publish only something which is already on MS. We have special checks which fail job if something is not on MS.

> "publish the MS marketplace version" - it actually doesn't but instead publishes the vsix found under the releases, correct?

Right, it is not allowed, so we have to build from releases, tags or commits around on last update date in MS. Please take into the account that we actually want to minimize extensions in this repo, so we will work more closer with extension authors to enable them publishing directly. Only unmaintained extensions or from MS should end up here. Most likely unmaintained should be removed also and published only once.

> Does this only apply to GitHub hosted repos?

Yes, but easily to extend to GitLab which we should do soon. We only query for latest release, otherwise we rely on local git for checking tags/commits. So most of code should be reused.

> Doesn't the remove of checkout likely break a bunch of extensions that have master/main as "dev snapshot" and a stable branch which was previously explicit configured via checkout? Is there a new way of doing the same?

We don't want to rely on it, since it is pretty hard. Rather push maintainers to provide proper releases and fail otherwise. That's said we look closer now what the publishing job is doing each night to learn.

akosyakov:
> My biggest point is: these are a lot of "design changes" with more effect to users than the move to Eclipse Foundation and different sign-up (which was the reason for some extension maintainers to not publish any more on their own or to not start that - for whatever reason), so I still miss an announcement.

Got it, give us a bit of time to sort out how good is a new approach. We maybe even need a commit again. We struggled with previous upgrade process which was tedious and yield not really good results.

PR #497, opened 2021-11-12
Cleanly install dependencies

filiptronicek:
Don't update the lockfile (`package-lock.json` or `yarn.lock`) when installing extension dependencies.

## How to test
The easiest way to test this is to run all the commands done by https://github.com/open-vsx/publish-extensions/blob/master/publish-extension.js manually.


bigint (approved):
LGTM :shipit:

filiptronicek:
We should do this in a separate PR if it proves useful.

PR #498, opened 2021-11-13
fixes to https://github.com/open-vsx/publish-extensions/pull/494

akosyakov:


types.d.ts
```diff
@@ -39,7 +39,19 @@ export interface PublishStat {
...
     hitMiss: {
-        [id: string]: (ExtensionStat | ExtensionStat) & { hit: boolean }
+        [id: string]: (ExtensionStat | ExtensionStat) & { hit: boolean }
     }
     failed: string[]
+}
+
+export interface Extension {
+    id: string,
+    repository?: string
```
akosyakov, line 49: should become mandatory eventually

types.d.ts
```diff
@@ -39,7 +39,19 @@ export interface PublishStat {
...
+    version?: string
+    checkout?: string
+    location?: string
+    prepublish?: string
+    download?: string
```
akosyakov, line 54: should be removed eventually

filiptronicek (approved):

PR #499, opened 2021-11-15
update extensions to the new format

filiptronicek:


akosyakov:
Should not it be a part of #494 that we can verify changes?

filiptronicek:
> Should not it be a part of #494 that we can verify changes?

@akosyakov I changed the PR to target #494, was a bit scared of having changes to extensions.json in the same PR, but we can merge this to get it there.

akosyakov:
I don’t think #494 is finished without changes to extensions.json, please merge it there

PR #501, opened 2021-11-16
vscode-gettext update

GitMensch:
dropping the explicit checkout to let further updates be auto-detected


GitMensch:
Typo... closed in favor of #502

PR #502, opened 2021-11-16
 vscode-gettext update

GitMensch:
dropping the explicit checkout to let further updates be auto-detected


filiptronicek:
I think this does not apply anymore, since we don't have versions in `extensions.json` as of last week.

PR #509, opened 2021-11-22
Add cup-lang.cup extension

jgrajewsky:


akosyakov:
@jgrajewsky Please consider to reach out to the extension author first. Please see [why it is recommended.](https://github.com/open-vsx/publish-extensions#when-to-add-an-extension).

The docs for publishing the extension are here: https://github.com/eclipse/openvsx/wiki/Publishing-Extensions. Technically this should be straightforward. You basically must get a token and call `ovsx publish` with it. For example, this is how RedHat publishes their java extension: redhat-developer/vscode-java@ff51dbf/Jenkinsfile#L73-L82

jgrajewsky:
I'm the author. Should I still follow [this](https://github.com/eclipse/openvsx/wiki/Publishing-Extensions)?

akosyakov:
@jgrajewsky yes, it would be the best. Let us know if for some reasons it does not work for you.

akosyakov:
Awesome!

PR #511, opened 2021-11-24
Allow for other git hosts other than GitHub

filiptronicek:
The only change here is that if the host of the `repository` property isn't GitHub.com, the script does all the logic locally, no API calling.

## How to test

Execute the following:
```
EXTENSIONS=pflannery.vscode-versionlens SKIP_PUBLISH=true FORCE=true GITHUB_TOKEN=blabla node publish-extensions.js
```
(`GITHUB_TOKEN` can really be `blabla`, we don't need to access the API, but we need to provide it.)

### DIFF
--- a/extensions.json
+++ b/extensions.json
@@ -973,7 +973,7 @@
     "repository": "https://github.com/pj3677/vscode-protobuf"
   },
   "pflannery.vscode-versionlens": {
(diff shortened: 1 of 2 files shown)


bigint:
So for github we are using api and for others just git clone right?

akosyakov:
> So for github we are using api and for others just git clone right?

Yes, in GitLab case there are not usable releases.

akosyakov (approved):
lgtm

PR #513, opened 2021-11-25
Check whether extension has an OSI approved license

### LINKED ISSUE #510: auto check that license is OSI approved
there is a package https://www.npmjs.com/package/osi-licenses It would be interesting to know whether it could work

filiptronicek:
Fixes #510

## How to test
1. Add the following to `extensions.json`
```json
  "filiptronicek.unlicensed": {
    "repository": "https://github.com/filiptronicek/test-unlicensed-repo"
  },
```
2. Execute this command and watch the terminal output (also add your own `GITHUB_TOKEN`, so that the script can request the GitHub API for the license)
```
GITHUB_TOKEN=xxx EXTENSIONS=filiptronicek.unlicensed node publish-extensions
```

## ToDo
- [ ] Make sure licenses are checked in every case, including fetching from GitHub releases
	- Using the GitHub API to check the license should be a last resort, because it is not possible to query a license for a specific time

### DIFF
--- a/lib/resolveExtension.js
+++ b/lib/resolveExtension.js
@@ -4,9 +4,9 @@ const fs = require('fs');
 const path = require('path');
 const Octokit = require('octokit').Octokit;
 const readVSIXPackage = require('vsce/out/zip').readVSIXPackage;
-const checkLicense = require('osi-license-checker');
 const download = require('download');
 const exec = require('./exec');
+const { checkLicense } = require('./checkLicense');
 
 const token = process.env.GITHUB_TOKEN;
 if (!token) {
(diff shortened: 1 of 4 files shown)


filiptronicek:
Published https://www.npmjs.com/package/osi-license-checker

lib/resolveExtension.js
```diff
@@ -53,6 +54,26 @@ exports.resolveExtension = async function ({ id, repository, location }, ms) {
   await exec(`git clone --filter=blob:none --recurse-submodules ${repository} ${repoPath}`, { quiet: true });
 
   const packagePath = [repoPath, location, 'package.json'].filter(p => !!p).join('/');
+
+  //#region Check if the extension has an OSI-approved open-source license
+  try {
```
akosyakov, line 59: Should not we check a revision form which we build vsix file?

filiptronicek: Great catch, @akosyakov, I was thinking about doing that, settled on "licenses don't change from commit to commit, right?". In edge cases where a license might be changed, added, or removed, I'll integrate it into the `resolveVersion` function to avoid otherwise duplicating the code over and over.

https://github.com/open-vsx/publish-extensions/blob/4d657302ecdb5f219159486335ec883df1216bf9/lib/resolveExtension.js#L60-L71

filiptronicek: Just looked into this again, and it seems there is no way to fetch the license for a specific time in the repository, just the current state. I checked with both the REST and GraphQL APIs, the only parameters allowed are `name` and `owner`.

PR #516, opened 2021-12-01
Add extension vemonet.stardog-rdf-grammars

vemonet:
Added the extension `vemonet.stardog-rdf-grammars` (for RDF language highlighting) in the `extensions.json` file

On OpenVSX: https://open-vsx.org/extension/vemonet/stardog-rdf-grammars
The source code: https://github.com/vemonet/stardog-vsc

### DIFF
--- a/extensions.json
+++ b/extensions.json
@@ -1162,6 +1162,9 @@
   "Unity.unity-debug": {
     "repository": "https://github.com/Unity-Technologies/vscode-unity-debug"
   },
+  "vemonet.stardog-rdf-grammars": {
+    "repository": "https://github.com/vemonet/stardog-vsc"
+  },
   "vhdlwhiz.vhdl-by-vhdlwhiz": {
     "repository": "https://github.com/jonasjj/awesome-vhdl"
   },
--- a/extensions.json
+++ b/extensions.json
@@ -1163,7 +1163,8 @@
     "repository": "https://github.com/Unity-Technologies/vscode-unity-debug"
   },
   "vemonet.stardog-rdf-grammars": {
-    "repository": "https://github.com/vemonet/stardog-vsc"
+    "repository": "https://github.com/vemonet/stardog-vsc",
+    "location": "stardog-rdf-grammars"
   },
   "vhdlwhiz.vhdl-by-vhdlwhiz": {
     "repository": "https://github.com/jonasjj/awesome-vhdl"


akosyakov:
@vermonent Please consider to reach out to the extension author first. Please see [why it is recommended.](https://github.com/open-vsx/publish-extensions#when-to-add-an-extension).

The docs for publishing the extension are here: https://github.com/eclipse/openvsx/wiki/Publishing-Extensions. Technically this should be straightforward. You get a token and call `ovsx publish` with it. For example, this is how RedHat publishes their java extension: [redhat-developer/vscode-java@ff51dbf/Jenkinsfile#L73-L82](https://github.com/redhat-developer/vscode-java/blob/5d4f8d58b8e919534800ca7bc41e8513c288f573/Jenkinsfile#L78-L82)

[Here](https://github.com/PeterWone/vsc-print/pull/121) is also an example PR to contribute release GH action which can publish to GitHub, MS and OpenVSX at the same time.

akosyakov:
@vemonet I just noticed that this extension belongs to you and you are already publishing it to OpenVSX. Why do you want to add it here? It is only for cases when author does not take responsibilities in OpenVSX. You are already do.

vemonet:
Hi @akosyakov, thanks for your reply, I wanted to be able to install this extension (Apache license) directly from the extension catalog of a Code server (https://github.com/cdr/code-server), I might be wrong but I thought it needed to be added to this `publish-extensions` list 

There is already an issue open for some time on the repo for the original authors, but I will ask again

akosyakov:
I see, you are maintaining a fork.

@filiptronicek Could you check please why `vemonet.stardog-rdf-grammars: failed to resolve`?

filiptronicek:
@akosyakov it failed because the location of the extension is not the root in the repo, rather a subdirectory. Fixed it in [`989f1fa` (#516)](https://github.com/open-vsx/publish-extensions/pull/516/commits/989f1fab85d131eab61dadfeb313244d8081d63d)

akosyakov:
Thank Filip,

my take on it that we should close this PR, since @vemonet already publishing a fork to OpenVSX and can maintain it. It would be better than we are doing it.

Instead we should work with original authors to enable them publishing to Open VSX.

vemonet:
I was going to close it, the extension is properly published to OpenVSX

The issue is due to code-server not indexing it, more info here: https://github.com/stardog-union/stardog-vsc/issues/81

PR #520, opened 2021-12-08
Add evilz.vscode-reveal

evilz:
Hello, 

I am the maintainer  of evilz.vscode-reveal and I just want to add it on https://open-vsx.org/

### DIFF
--- a/extensions.json
+++ b/extensions.json
@@ -1302,5 +1302,8 @@
   },
   "zxh404.vscode-proto3": {
     "repository": "https://github.com/zxh0/vscode-proto3"
+  },
+  "evilz.vscode-reveal": {
+    "repository": "https://github.com/evilz/vscode-reveal"
   }
 }


akosyakov:
@evilz It would be better if you take the ownership and publish yourself. Please see [why it is recommended.](https://github.com/open-vsx/publish-extensions#when-to-add-an-extension).

This repo is to seed the registry with extensions when maintainers cannot, for instance because they work for MS, or not active anymore. 

The docs for publishing the extension are [here](https://github.com/eclipse/openvsx/wiki/Publishing-Extensions). Technically this should be straightforward. You get a token and call `ovsx publish` with it. For example, this is how RedHat publishes their java extension: [redhat-developer/vscode-java@ff51dbf/Jenkinsfile#L73-L82](https://github.com/redhat-developer/vscode-java/blob/5d4f8d58b8e919534800ca7bc41e8513c288f573/Jenkinsfile#L78-L82)

Let me know if you are opened to a PR contributing a release GH action both to MS marketplace and OpenVSX, similar to https://github.com/PeterWone/vsc-print/pull/121.

BTW at Gitpod we support open source developers by providing unlimited access to platform. Feel free to reach out via https://www.gitpod.io/docs/professional-open-source

PR #525, opened 2021-12-18
fix build of `johnsoncodehk.volar` & `johnsoncodehk.vscode-typescript-vue-plugin`

anbraten:
Problem:
- command `yarn compile` has to be `yarn build`
- `yarn install` needs to be executed in root folder for the volar repo as it is a yarn / lerna monorepo. 
  `"prepublish": "yarn install && ...",` works as the prepublish command is currently ignoring the `location` option and by that it is executed in the repo root: https://github.com/open-vsx/publish-extensions/blob/377e1d79b7d0a7d307df34b3ae9f30e86b3ccd2c/publish-extension.js#L63

### DIFF
--- a/extensions.json
+++ b/extensions.json
@@ -559,12 +559,12 @@
   },
   "johnsoncodehk.volar": {
     "repository": "https://github.com/johnsoncodehk/volar",
-    "prepublish": "yarn compile",
+    "prepublish": "yarn install && yarn build",
     "location": "extensions/vscode-vue-language-features"
   },
   "johnsoncodehk.vscode-typescript-vue-plugin": {
     "repository": "https://github.com/johnsoncodehk/volar",
-    "prepublish": "yarn compile",
+    "prepublish": "yarn install && yarn build",
     "location": "extensions/vscode-typescript-vue-plugin"
   },
   "jolaleye.horizon-theme-vscode": {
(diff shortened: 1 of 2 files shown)


akosyakov:
@johnsoncodehk it would be really great if you can take over publishing and ownership. 🙏  Please see [why it is recommended.](https://github.com/open-vsx/publish-extensions#when-to-add-an-extension).

The docs for publishing the extension are [here](https://github.com/eclipse/openvsx/wiki/Publishing-Extensions). Technically this should be straightforward. You get a token and call `ovsx publish` with it. For example, this is how RedHat publishes their java extension: [redhat-developer/vscode-java@ff51dbf/Jenkinsfile#L73-L82](https://github.com/redhat-developer/vscode-java/blob/5d4f8d58b8e919534800ca7bc41e8513c288f573/Jenkinsfile#L78-L82)

[Here](https://github.com/PeterWone/vsc-print/pull/121) is also an example PR to contribute release GH action which can publish to GitHub, MS and OpenVSX at the same time.

@filiptronicek can help to provide a PR for your extensions to automate publishing, so we can remove them from here.

filiptronicek (approved):
The changes LGTM.

BTW thanks for fixing the typo 😄

PR #526, opened 2021-12-18
editorconfig: change id to lower case

6543:


6543:
@filiptronicek do you have 3sec time - it should be a simple fix after all :)

filiptronicek:
@6543 it looks like indeed in the README of the repo the ID is lowercase, but on the Microsoft Marketplace, it is capitalized, which needs to be the same because of version resolutions. 

image

PR #534, opened 2022-01-12
Add `adrieankhisbe.vscode-ndjson`

johnwchadwick:
I made an effort to contact the owner in December via e-mail, but I haven't gotten a response. I can't open an issue in the repository in question because it does not have issues enabled. Because this project is under an OSI-approved license (MIT) I think it is suitable for inclusion here. I ran the publish script with this in GitPod and was able to install the extension and it seems to work.


filiptronicek:
Hi @johnwchadwick

PR #548, opened 2022-02-17
Use the xml vsix manifest as a primary data source

filiptronicek:
This fixes license resolutions for extensions like https://github.com/Microsoft/vscode-java-test, which don't have a `license` field in their manifest (`package.json`), but do as a file in the repository, while offering `vsix` release assets at the same time.

## How to test

```
EXTENSIONS=vscjava.vscode-java-test SKIP_PUBLISH=true FORCE=true GITHUB_TOKEN=YOUR_PAT_HERE node publish-extensions
```

The above should run successfully, you can try any other extension instead of the Java one and to do more at the same time comma-separate them.

### DIFF
--- a/publish-extension.js
+++ b/publish-extension.js
@@ -103,7 +103,7 @@ const { createVSIX } = require('vsce');
             }
         }
         // TODO(ak) check license is open-source
-        if (!xmlManifest?.PackageManifest?.Metadata[0]?.License && !(packagePath && await ovsx.isLicenseOk(packagePath, manifest))) {
+        if (!xmlManifest?.PackageManifest?.Metadata[0]?.License?.[0] && !(packagePath && await ovsx.isLicenseOk(packagePath, manifest))) {
             throw new Error(`${extension.id}: license is missing`);
         }
 


filiptronicek:
@jeanp413 added it now, thanks for noticing

publish-extension.js
```diff
@@ -103,7 +103,7 @@ const { createVSIX } = require('vsce');
             }
         }
         // TODO(ak) check license is open-source
-        if (!manifest?.license && !(packagePath && await ovsx.isLicenseOk(packagePath, manifest))) {
+        if (!xmlManifest?.PackageManifest?.Metadata[0]?.License && !(packagePath && await ovsx.isLicenseOk(packagePath, manifest))) {
```
jeanp413, line 106: License [is an array ](https://github.com/microsoft/vscode-vsce/blob/228d0e1fc35031786d1fa436e05cb091ffa8a430/src/xml.ts#L17)so better check if it has at least one element 
```suggestion
        if (!xmlManifest?.PackageManifest?.Metadata[0]?.License?.[0] && !(packagePath && await ovsx.isLicenseOk(packagePath, manifest))) {
```

jeanp413 (changes requested):

publish-extension.js
filiptronicek: Thanks!

jeanp413 (approved):
LGTM

PR #549, opened 2022-02-18
Add: VSCode Cheatsheet

tennox:
https://github.com/MattiasPernhult/vscode-cheatsheet

MIT license:
https://github.com/MattiasPernhult/vscode-cheatsheet/blob/master/LICENSE

- [x] I have tried [reaching out to the extension maintainers](https://github.com/MattiasPernhult/vscode-cheatsheet/issues/1) about publishing this extension to OpenVSX (if no, please create an issue in the extension's repo using [this template](https://github.com/open-vsx/publish-extensions/blob/HEAD/docs/external_contribution_request.md).

## Description

This PR introduces X change for Y reason.

### DIFF
--- a/extensions.json
+++ b/extensions.json
@@ -669,6 +669,9 @@
   "mathiasfrohlich.Kotlin": {
     "repository": "https://github.com/mathiasfrohlich/vscode-kotlin"
   },
+  "MattiasPernhult.vscode-cheatsheet": {
+    "repository": "https://github.com/MattiasPernhult/vscode-cheatsheet"
+  },
   "max-SS.Cyberpunk": {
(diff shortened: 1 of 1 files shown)


filiptronicek (commented):
Hey @tennox, thanks for contributing this extension. We should probably mention the developer in the issue or message him by other means: the extension is from 2016 and I have tried for a while now to get it working, but it seems pretty tricky. If we could get the `vsix` of the package extension, we could just publish that and not have to worry about keeping it in the repo, because it looks like 6 years without an update sounds like a good reason to suspect it's not actively being worked on 😄.

tennox:
After using the extension for a bit... not worth it :sweat_smile: 

Thanks for the effort :+1:

PR #552, opened 2022-03-01
Add a `custom` property

filiptronicek:
This PR introduces a `custom` property available to extensions to let them have an entirely customized process, because not all extensions are built the same, meaning even some of the biggest ones we have are failing.

This PR adds a very powerful option, in docs we should make it clear that this should be used as a last measure (we need to favor extensions with no setup, if they require some then we should point to the `prepublish` setting).

## How to test

Try to execute the following to test that everything is working correctly:

```
EXTENSIONS=formulahendry.auto-rename-tag FORCE=true GITHUB_TOKEN=YOUR_PAT node publish-extensions
```

### DIFF
--- a/extensions-schema.json
+++ b/extensions-schema.json
@@ -24,6 +24,10 @@
         "type": "string",
         "description": "Relative path of the extension vsix file inside the git repo (i.e. when it is built by prepublish commands"
       },
+      "custom": {
+        "type": "string",
+        "description": "Build using a custom script. Must output an `extension.vsix` file in the `location` directory."
+      },
       "timeout": {
         "type": "number",
         "description": "Timeout to build the extension vsix from sources."
--- a/extensions-schema.json
+++ b/extensions-schema.json
@@ -25,7 +25,7 @@
         "description": "Relative path of the extension vsix file inside the git repo (i.e. when it is built by prepublish commands"
       },
       "custom": {
-        "type": "string",
+        "type": "array",
         "description": "Build using a custom script. Must output an `extension.vsix` file in the `location` directory."
       },
       "timeout": {
--- a/extensions.json
+++ b/extensions.json
@@ -116,9 +116,6 @@
   "ban.spellright": {
     "repository": "https://github.com/bartosz-antosik/vscode-spellright"
   },
-  "batisteo.vscode-django": {
-    "repository": "https://github.com/vscode-django/vscode-django"
-  },
   "BazelBuild.vscode-bazel": {
     "repository": "https://github.com/bazelbuild/vscode-bazel"
   },
@@ -520,9 +517,6 @@
   "itsjonq.owlet": {
     "repository": "https://github.com/ItsJonQ/owlet"
   },
-  "itsmajestix.vscode-ti": {
-    "repository": "https://github.com/ItsMajestiX/VSCode-Ti"
-  },
   "ivangabriele.vscode-git-add-and-commit": {
     "repository": "https://github.com/ivangabriele/vscode-git-automator-legacy"
   },
@@ -630,9 +624,6 @@
   "kortina.vscode-markdown-notes": {
     "repository": "https://github.com/kortina/vscode-markdown-notes"
   },
-  "kosz78.nim": {
-    "repository": "https://github.com/pragmagic/vscode-nim"
-  },
   "laradump.laradump": {
     "repository": "https://github.com/TheJenos/laradump-vscode"
   },
@@ -653,7 +644,8 @@
   },
   "mads-hartmann.bash-ide-vscode": {
     "repository": "https://github.com/bash-lsp/bash-language-server",
-    "location": "vscode-client"
+    "location": "vscode-client",
+    "prepublish": "yarn"
   },
   "magicstack.MagicPython": {
     "repository": "https://github.com/MagicStack/MagicPython"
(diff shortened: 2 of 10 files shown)


akosyakov:
What is the motivation?

filiptronicek:
@akosyakov mostly testing, but it should fix publishing some harder Microsoft extensions, specifically https://github.com/microsoft/vscode-js-profile-visualizer, because it requires the deps install to never occur, because if you do `npm install` in the subfolder of the cloned repo it fails. I tried going around it yesterday but it seems pretty tricky, hence this. I'm sure it will be also helpful for other extensions as well, but if we can fix these exts without such an aggressive option, I'm pretty sure it would be better.

akosyakov:
I just wonder we already have `prepublish` and `extensionFile` and we can change location should not it be enough?

extensions.json
```diff
@@ -881,7 +882,7 @@
   "ms-vscode.vscode-js-profile-table": {
     "repository": "https://github.com/microsoft/vscode-js-profile-visualizer",
     "location": "packages/vscode-js-profile-table",
-    "prepublish": "npm run compile"
+    "custom": "npm install && npm run compile && cd packages/vscode-js-profile-table && npx vsce package --yarn --out extension.vsix"
```
manuth, line 885: For the sake of readability and maintainability, I'd love to be able to optionally pass an array to `custom`.
In case of the `ms-vscode.vscode-js-profile-table`, your `custom` might look like this:

```json
"custom": [
    "npm install",
    "npm run compile",
    "cd packages/vscode-js-profile-table && npx vsce package --yarn --out extension.vsix",
    "pwd", "// Prints ${context.repo}"
]
```

This might be more readable especially when working with extensions which require a lot of commands being executed for building.

manuth (commented):
Looks awesome to me - I'm looking forward to have the profiler (and stuff like jupyter) running in Codium at some point! 😄 

There's one small change I'd like to see in this PR

extensions.json
filiptronicek: Oh, that's a magnificent idea! It really would make it more readable and it would make total sense to do it. Will take a look this week / early next week.

extensions-schema.json
```diff
@@ -25,7 +25,7 @@
         "description": "Relative path of the extension vsix file inside the git repo (i.e. when it is built by prepublish commands"
       },
       "custom": {
-        "type": "string",
+        "type": "array",
```
manuth, line 28: I'd recommend you to allow both strings and arrays (if this wouldn't look too hacky):

```json
"custom": {
    "description": "Build using a custom script. Must output an `extension.vsix` file in the `location` directory.",
    "oneOf": [
        {
            "type": "string",
            "default": ""
        },
        {
            "type": "array",
            "default": []
        }
    ]
}
```

extensions.json
```diff
@@ -874,7 +879,12 @@
   "ms-vscode.vscode-js-profile-table": {
     "repository": "https://github.com/microsoft/vscode-js-profile-visualizer",
     "location": "packages/vscode-js-profile-table",
-    "custom": "npm install && npm run compile && cd packages/vscode-js-profile-table && npx vsce package --yarn --out extension.vsix" 
+    "custom": [
+      "npm install",
+      "npm run compile",
+      "cd packages/vscode-js-profile-table",
+      "npx vsce package --yarn --out extension.vsix"
```
manuth, line 886: If you were to reset the cwd for each command (which I'd recommend you to do), you'd have to chain these two commands with a `&&`:

```suggestion
      "cd packages/vscode-js-profile-table && npx vsce package --yarn --out extension.vsix"
```

publish-extension.js
```diff
@@ -41,7 +41,9 @@ const { createVSIX } = require('vsce');
             await exec(`git checkout ${context.ref}`, { cwd: context.repo });
             if (extension.custom) {
                 try {
-                    await exec(extension.custom, { cwd: context.repo });
+                    for (const command of extension.custom) {
+                        await exec(command, { cwd: context.repo });
+                    }
```
manuth, line 46: I'd recommend to allow both strings and arrays:
```suggestion
                    const commands = typeof extension.custom === "string" ? [extension.custom] : extension.custom;

                    for (const command of commands) {
                        await exec(command, { cwd: context.repo });
                    }
```

types.d.ts
```diff
@@ -58,7 +58,7 @@ export interface Extension {
     location?: string
     prepublish?: string
     extensionFile?: string
-    custom?: string
+    custom?: string[]
```
manuth, line 61:
```suggestion
    custom?: string | string[]
```

manuth (commented):
Hope you don't mind me interrupting once more. I think it might be a good idea to let the devs choose whether to provide a `custom` array or a string.

Looks great 😄

extensions-schema.json
filiptronicek: I don't have a problem with adding this, but I am not sure if it is something that useful. If we stick to one format it looks to me like it does not have any disadvantages as opposed to an array, maybe just that a single command would be smaller in size. @manuth do you think that there could be any differences in the two approaches from how the commands are executed?

manuth: No, not really - I think we'll be fine without this functionality!
I'm perfectly fine without it 😄

extensions.json
filiptronicek: Good catch, thank you!

PR #554, opened 2022-03-02
Add Jupyter extensions

filiptronicek:
This PR adds missing Microsoft Jupyter extensions.

### DIFF
--- a/extensions.json
+++ b/extensions.json
@@ -785,7 +785,13 @@
     "repository": "https://github.com/microsoft/vscode-jupyter-keymap"
   },
   "ms-toolsai.jupyter-renderers": {
-    "repository": "https://github.com/Microsoft/vscode-notebook-renderers"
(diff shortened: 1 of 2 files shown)


filiptronicek:
`ms-toolsai.jupyter-renderers` is sooo flustrating to publish: I can package it up using the provided commands, but it's so hard to do: I only achieved success while using Node 14 (Node 16 requires more config). If we ever want to just publish a version of the extension pre-packaged, we can do so with [ms-notebook-renderers.vsix.zip](https://github.com/open-vsx/publish-extensions/files/8173477/ms-notebook-renderers.vsix.zip) [GitHub doesn't allow `.vsix` file uploads].

filiptronicek:
Linking https://github.com/microsoft/vscode-notebook-renderers/issues/96 here because it might help with figuring out how to package it properly.

PR #556, opened 2022-03-08
Add reporting for most popular Marketplace Extensions

filiptronicek:
## Description

This PR adds a script which reports stats about the most popular extensions from MS which we don't have on OpenVSX, maybe pointing us to extensions which we don't have from Microsoft or which we should ask their maintainers about.

It also adds docs about understanding the reports that the GitHub Action generates as well as some quality of life improvements in the publishing and reporting scripts.

## How to test

- For the docs part, check it out and read through it
- For the reporting part, check the [workflow run](https://github.com/open-vsx/publish-extensions/actions/runs/2383294147) to see what it produces

### DIFF
--- a/docs/understanding_reports.md
+++ b/docs/understanding_reports.md
@@ -1,12 +1,12 @@
-## Understanding statistics
+# Understanding statistics
 
 This document is here to help you understand reports from [the nightly publishing job](https://github.com/open-vsx/publish-extensions/actions/workflows/publish-extensions.yml).
 
-### How to gather these statistics
+## How to gather these statistics
 
 If you click on a job in the GitHub Actions tab, there is an `Artifacts` section at the bottom of the page, from which you can download the `report`, which after unarchiving reveals two files: `result.log` and `stat.json`.
 
-### `stat.json`
+## `stat.json`
 
 This is the machine-readable data that the next file - `result.log` is generated from. In it, you can find 9 different categories of extensions:
 - `upToDate` - these extensions are the extensions, which have the same version published to OpenVSX as well as the Microsoft Marketplace.
@@ -19,11 +19,11 @@ This is the machine-readable data that the next file - `result.log` is generated
 - `hitMiss`
 - `resolutions` is a list of all extensions and the way they have been resolved: `latest`, `matchedLatest`. `releaseTag`, `tag` or `releaseAsset`.
 
-### `result.log`
+## `result.log`
 
 This file is the one that should provide a quick overview of how the repository is doing. It has many percentages, numbers and sections, so that you can quickly take a look and get the information you want. These are mostly made from the `stat.json` file and pretty self-explanatory, but there are some that are a big more comples:
 
-#### `Weighted publish percentage`
+### `Weighted publish percentage`
 
 This metric's goal is to provide the one number you need to see if the big and most used extensions are up-to-date and existing on OpenVSX. This value is computed as follows (in pseudo-code):
 ```ts
@@ -32,3 +32,7 @@ const totalInstalls = sum(upToDate); // a sum of all install from the Microsoft 
 
 const weightedPublishPercentage = upToDateInstalls / totalInstalls;
 ```
+
+### Microsoft-published extensions
+
+In the `Summary`, you can see a how many extensions published by Microsoft are defined in our repo, how many of them are outdated and how many of them are unstable. For further details with all of the failing extensions, refer to the `MS extensions` section in the file. Under there, you can find more details, like the specific IDs of the extensions and their install counts. 
--- a/docs/understanding_reports.md
+++ b/docs/understanding_reports.md
@@ -4,7 +4,7 @@ This document is here to help you understand reports from [the nightly publishin
 
 ## How to gather these statistics
 
-If you click on a job in the GitHub Actions tab, there is an `Artifacts` section at the bottom of the page, from which you can download the `report`, which after unarchiving reveals two files: `result.log` and `stat.json`.
+If you click on a job in the GitHub Actions tab, there is an `Artifacts` section at the bottom of the page, from which you can download the `report`, which after unarchiving reveals three files: `result.log`, `stat.json` and `meta.json`.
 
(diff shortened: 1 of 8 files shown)


docs/understanding_reports.md
```diff
@@ -0,0 +1,38 @@
...
+
+This is the machine-readable data that the next file - `result.log` is generated from. In it, you can find 9 different categories of extensions:
+- `upToDate` - these extensions are the extensions, which have the same version published to OpenVSX as well as the Microsoft Marketplace.
+- `outdated` are all of the extensions, which have versions on OpenVSX, which are behind the ones on the Microsoft Marketplace. 
+- `unstable` is the category, which has all extensions, that are most likely incorrectly published from the extensions' nightly or beta builds, since on the Microsoft Marketplace has a version which is smaller than the one on OpenVSX.
+- `notInOpen` includes extensions that simply failed to ever be published to OpenVSX, which means they should get special attention - fix them or remove them :)
+- `notInMs` - extensions that aren't published on the Microsoft Marketplace
+- `failed` - the extensions that for some reason failed with their publishing.
+- `msPublished` - all extensions published by Microsoft Corporation.
+- `hitMiss`
```
loujaybee, line 19: Are we missing a description? 😁

filiptronicek: Yes we are, I need to figure out what this thing does 😅

filiptronicek: Fixed in https://github.com/open-vsx/publish-extensions/pull/556/commits/78632d51ba4b9fda2e1ab4b97d6ffaddeddfe39b :)

docs/understanding_reports.md
```diff
@@ -0,0 +1,40 @@
...
+This is the machine-readable data that the next file - `result.log` is generated from. In it, you can find 9 different categories of extensions:
+
+- `upToDate` - these extensions are the extensions, which have the same version published to OpenVSX as well as the Microsoft Marketplace.
+- `outdated` are all of the extensions, which have versions on OpenVSX, which are behind the ones on the Microsoft Marketplace.
+- `unstable` is the category, which has all extensions, that are most likely incorrectly published from the extensions' nightly or beta builds, since on the Microsoft Marketplace has a version which is smaller than the one on OpenVSX.
+- `notInOpen` includes extensions that simply failed to ever be published to OpenVSX, which means they should get special attention - fix them or remove them :)
+- `notInMs` - extensions that aren't published on the Microsoft Marketplace
+- `failed` - the extensions that for some reason failed with their publishing.
+- `msPublished` - all extensions published by Microsoft Corporation.
+- `hitMiss` - extensions which, in <abbr title="Month-To-Date">MTD</abbr>, have been updated on OpenVSX within 2 days after the Microsoft Marketplace.
```
filiptronicek, line 20: Sadly, folks reading these docs on GitHub will not have this abbreviation displayed to them, because this is not yet something GitHub does [[1](https://github.com/github/feedback/discussions/9681)]

docs/understanding_reports.md
```diff
@@ -0,0 +1,40 @@
+# Understanding statistics
+
+This document is here to help you understand reports from [the nightly publishing job](https://github.com/open-vsx/publish-extensions/actions/workflows/publish-extensions.yml).
+
+## How to gather these statistics
+
+If you click on a job in the GitHub Actions tab, there is an `Artifacts` section at the bottom of the page, from which you can download the `report`, which after unarchiving reveals two files: `result.log` and `stat.json`.
```
andreafalzetti, line 7: do you also want to mention the fact that `report.zip` also contains `meta.json`?

report-extensions.js
```diff
@@ -114,16 +156,25 @@ function sortedKeys(s) {
...
     summary += `Not in MS marketplace: ${notInMS} (${(notInMS / total * 100).toFixed(0)}%)\r\n`;
-    summary += `Failed to publish: ${stat.failed.length} (${(stat.failed.length / total * 100).toFixed(0)}%) \r\n`
-    summary += `MS is publisher: ${msPublished} (${(msPublished / total * 100).toFixed(0)}%)\r\n`;
-    summary += `\r\n`;
+    summary += `Failed to publish: ${stat.failed.length} (${(stat.failed.length / total * 100).toFixed(0)}%) \r\n`;
+    summary += `Microsoft:\r\n`;
+    summary += `Total: ${msPublished} (${(msPublished / total * 100).toFixed(0)}%)\r\n`;
+    summary += `Outdated: ${msPublishedOutdated.length}\r\n`;
+    summary += `Unstable: ${msPublishedUnstable.length}\r\n`;
+    summary += `Missing: ${missingMs.length} (we could publish ${couldPublishMs.length} out of that)`
```
andreafalzetti, line 177:
```suggestion
    summary += `Missing: ${missingMs.length} (we could publish ${couldPublishMs.length} out of that)\r\n`
```

as it currently reads as:

```
Missing: 13 (we could publish 0 out of that)Total resolutions: 387
```

report-extensions.js
```diff
@@ -143,61 +194,85 @@ function sortedKeys(s) {
 
     let content = summary;
     if (outdated) {
-        if (process.env.VALIDATE_PR !== 'true') {
-            process.exitCode = -1;
```
akosyakov, line 147: why don't we fail anymore?

it seems that not all extensions are up to date, but job is green one, it looks misleading

filiptronicek: Thanks for bringing this up, @akosyakov. I see I forgot to mention it in the PR description, but the change was made because of the fact that the automatic job is always failing, and likely we will never hit green with all extensions, and even if we do, I would say it wouldn't last long. This is why now it is green by default, and fails if a big extension(s) that we have fails to publish (sum of all failing must be 5% of all total install counts). This means we can be alerted by a failing job about something going wrong which should always be prioritized (doesn't fail if an extension with 5 downloads fails to publish).

I understand that it looks misleading, but I believe that it's way more helpful than always seeing a red status indicator every time. I too recognize that it would be ideal if we fixed all of the extensions here but I do not think this is realistic currently.

PR #563, opened 2022-03-23
Remove kodetech-extensions

RobDangerous:
This PR removes all extensions from the kodetech-organization. I would like to take care of those myself from now on (see https://github.com/EclipseFdn/open-vsx.org/issues/895).


filiptronicek (approved):
@RobDangerous thanks for taking the time 🧡!

PR #566, opened 2022-03-25
Abort Publish Process if a Command Fails

manuth:
## Description
Currently, the `custom` property accepts a set of commands which are executed in order.
If an error occurrs, the commands which follow will still be executed. I think it might be a good idea to prevent extensions with failing commands form being published because this could lead to potentially broken extensions.

Changes made in this PR will make sure that the exit code equals 0. If the exit code doesn't equal 0, the `stdout` and `stderr` will be dumped and the publish-process of the corresponding extension will be aborted.

### DIFF
--- a/lib/exec.js
+++ b/lib/exec.js
@@ -14,7 +14,7 @@ const cp = require('child_process');
 /**
  * @param {string} command
  * @param {{cwd?: string, quiet?: boolean}} [options]
- * @returns {Promise<{ stdout: string, stderr: string }>}
+ * @returns {Promise<{ stdout: string, stderr: string, exitCode: number }>}
  */
 module.exports = async (command, options) => {
   if (!options?.quiet) {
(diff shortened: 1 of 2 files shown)


filiptronicek (commented):
Thanks @manuth greatly for the contribution!

I just checked, and it seems that (this may not be the case for all errors) the `exec` promise rejects on an error [[1](https://github.com/open-vsx/publish-extensions/blob/master/lib/exec.js#L33)], so I deliberately malformed a `custom` command on an extension and the script failed, never reaching the `throw` block you added, which means the functionality is already there as it seems: on an error in one of the commands the extension doesn't get processed.

I think this may not work for all use-cases, so I'd love it if you could share the ones it solves for you :) (I would guess some in which there is no `error` but the exit code is not 0)

manuth:
Oh - sorry, that's my bad. I thought, error might be related to something like a file not being executable or not being found.
It seems I should have read the docs 😂 

Sadly, I don't remember which of all the `Fix Extensions` workflow runs it was but in one of them, a vsxi-file got packaged even though the build process reported errors. I thought this might have been the root of the issue. But reading `Child process` docs I agree with you and think it might be best to close this PR.

PR #571, opened 2022-03-31
README.md: give publish-extensions.js a token

SamB:
For some reason, publish-extensions.js dies without a `GITHUB_TOKEN` environment variable,
even though the value seems to be irrelevant (in `SKIP_PUBLISH` mode).


filiptronicek:
Hey there @SamB 👋 , thanks for your contribution. The `GITHUB_TOKEN` is always needed, because it is not used for publishing, rather, it is used for resolving every extension and its releases, licenses etc. via octokit.js.

I will close the PR and add a note about it in the README, thanks for mentioning this 🙂.

PR #575, opened 2022-04-06
Remove Dart-Code extensions which are being published manually

DanTup:
I'm the author of the Dart/Flutter extensions and publishing these to open-vsx manually. This script published an update before I did yesterday, which results in it showing as "unverified" (whereas those published by me are verified).

I intend to continue publishing these extensions for the foreseeable future, so it would be better if this script doesn't beat me to it :-)

See https://github.com/EclipseFdn/open-vsx.org/issues/945.


filiptronicek (approved):
Thank you @DanTup! We really appreciate it.

PR #579, opened 2022-04-12
Add prepublish for python extension

jeanp413:
## Description
Add prepublish script for python extension needed for new versions


filiptronicek (approved):
TY JP!

PR #583, opened 2022-04-27
Fix `justusadam.language-haskell`

re-xyr:
- [x] I have tried reaching out to the extension maintainers about publishing this extension to OpenVSX (if not, please create an issue in the extension's repo using [this template](https://github.com/open-vsx/publish-extensions/blob/HEAD/docs/external_contribution_request.md)).

## Description

This PR adds `make all` to the custom build script of `justusadam.language-haskell` so that grammar files are correctly built.

### DIFF
--- a/extensions.json
+++ b/extensions.json
@@ -555,6 +555,7 @@
     "custom": [
       "[ $VERSION = '3.6.0' ] && git checkout 0475f2dace08a0b9f419c7521475564fa89f130f || true",
       "npm install",
+      "make all",
       "vsce package -o extension.vsix"
     ]
   },


filiptronicek:
Thank you @re-xyr for the contribution. As the extension build is succeeding, does that mean the extension is missing some functionality or is broken on OpenVSX? Just trying to understand what went wrong here 🙂

filiptronicek:
@re-xyr is this issue present in [v3.3.0](https://open-vsx.org/extension/justusadam/language-haskell/3.3.0) published to OpenVSX as well?

filiptronicek (approved):
Changes LGTM, unpublishing of the latest version will happen soon, hopefully, so that we can republish it with the fix.

manuel-plavsic:
Has this issue been fixed? I still get no syntax highlighting with open-vsx 3.6.0 (although 3.3.0 is fine).

re-xyr:
I think we'll need to wait for @filiptronicek to request a re-publish of 3.6.0 again, due to #589.

filiptronicek:
@re-xyr thanks for bringing this to my attention. I thought that #589 was just an improvement so that the extension build time would shrink? The last published version of 3.6.0 should be the one published after this PR has landed.

re-xyr:
Sadly no; in fact the build *timed out* on the CI machine ([see the build log here](https://github.com/open-vsx/publish-extensions/runs/6260461132?check_suite_focus=true#step:8:100)) so only a portion of the files was built, and eventually it re-uploaded a broken version. #589 hopefully fixes that.

manuel-plavsic:
@filiptronicek thanks for publishing the new version! The plugin seems to work correctly now, however, I assume that VSCodium still downloads the wrong version, because the same bugs still persists when installing the extension directly from VSCodium.

More in detail, even if I:

1. Delete both extensions `justusadam.language-haskell-3.6.0` and `haskell.haskell-2.2.0` from the location where the extensions are installed.
2. Reinstall the haskell plugin.
 
What happens is that the installed version of the syntax highlighting extension seems to be the right one (3.6.0, released 8 hours ago), however, the bug is still there.

However, I found a way to ensure the state-of-the-art version of the extension gets installed correctly:

1. Delete both existing "haskell" and "haskell syntax highlighting" extensions, i.e., step 1 above.
2. Manually download the VSIX https://open-vsx.org/extension/justusadam/language-haskell/3.6.0.
3. Proceed by installing the "haskell syntax highlighting" extension through the "Install from VSIX..." option.
4. install the haskell plugin directly from VSCodium, i.e., a search within VSCodium will be enough, ... no need to download the VSIX manually.

Does this behavior show also on your machines? If yes, I guess that a 3.7 version of the syntax highlighting plugin will prevent VSCodium from downloading a corrupted version.

re-xyr:
I tested on Gitpod, where it works. I'll test on VSCodium later.

PR #594, opened 2022-05-31
Add UNSAVED FILES plugin

kenneth-marut-work:
This plugin is [MIT licensed](https://marketplace.visualstudio.com/items/esamatti.unsaved/license) and has not been updated for some time but some of our users would like to have access to this functionality in our Theia-based IDE

### DIFF
--- a/extensions.json
+++ b/extensions.json
@@ -296,6 +296,9 @@
     "repository": "https://github.com/erlang-ls/vscode",
     "prepublish": "npm run compile"
   },
+  "esamatti.unsaved": {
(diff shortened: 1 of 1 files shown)


kenneth-marut-work:
Hi @filiptronicek. You're right, I did miss the previous comment earlier last year, my apologies. To save us both some time, I don't believe we need this extension anymore as we've found a workaround, so I would be fine if you close this issue. Thanks!

filiptronicek:
Will do, thanks!

PR #599, opened 2022-06-06
Update language packs

filiptronicek:
Don't publish the packs that are no longer available and publish the new pseudo language pack.

### DIFF
--- a/extensions.json
+++ b/extensions.json
@@ -688,10 +688,6 @@
   "ms-azuretools.vscode-docker": {
     "repository": "https://github.com/microsoft/vscode-docker"
   },
-  "MS-CEINTL.vscode-language-pack-bg": {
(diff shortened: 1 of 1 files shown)


jeanp413 (approved):

jeanp413:
@filiptronicek is it possible to use the vsix from github release instead of building it ourselves by removing the `"location": "i18n/vscode-language-pack-bg"` for example

filiptronicek:
@jeanp413 yes, it should be all resolving from GH releases (see [[1](https://github.com/open-vsx/publish-extensions/runs/6759140943?check_suite_focus=true#step:7:17)] in the PR check). The `location` is only used as a fallback, if the release didn't have a release page or there wouldn't be any assets attached.

PR #666, opened 2023-02-27
Fix `ms-toolsai.jupyter` extension node version

### LINKED ISSUE #664: Node.js nvmrc Issues with packaging the Jupyter Extension in Actions
The ms-toolsai.jupyter package is failing to package because node v14 is being used. The [.nvmrc](https://github.com/nvm-sh/nvm#nvmrc) file is properly specifying the necessary version and `nvm install` is run, but it is never used while the script runs. This leads to the following error because it is not using the proper node version.

```
[04:01:19] 'webpack-renderers' errored after 30 s
[04:01:19] TypeError: stripVTControlCharacters is not a function
    at /tmp/repository/main/gulpfile.js:285:24
    at Array.map (<anonymous>)
    at buildWebPack (/tmp/repository/main/gulpfile.js:285:10)
    at processTicksAndRejections (internal/process/task_queues.js:95:5)
    at async buildWebPackForDevOrProduction (/tmp/repository/main/gulpfile.js:199:9)
    at async /tmp/repository/main/gulpfile.js:211:5
[04:01:19] 'webpack' errored after 30 s
[04:01:19] 'prePublishBundle' errored after 30 s
```

jared-nussbaum-frbb:
- [x] I have tried reaching out to the extension maintainers about publishing this extension to Open VSX (if not, please create an issue in the extension's repo using [this template](https://github.com/open-vsx/publish-extensions/blob/HEAD/docs/external_contribution_request.md)).

## Description

This builds the `ms-toolsai.jupyter` extension with the proper version of node specified in the [.nvmrc](https://github.com/microsoft/vscode-jupyter/blob/main/.nvmrc)
Closes #664

### DIFF
--- a/extensions.json
+++ b/extensions.json
@@ -797,10 +797,7 @@
   "ms-toolsai.jupyter": {
     "repository": "https://github.com/microsoft/vscode-jupyter",
     "custom": [
-      "npm ci",
-      "python3 -m pip --disable-pip-version-check install -t ./pythonFiles/lib/python --no-cache-dir --implementation py --no-deps --upgrade -r requirements.txt --no-user",
-      "npm run clean",
-      "npm run package"
+      "source ~/.nvm/nvm.sh && nvm use && nvm install-latest-npm && npm install --global @vscode/vsce && npm ci && python3 -m pip --disable-pip-version-check install -t ./pythonFiles/lib/python --no-cache-dir --implementation py --no-deps --upgrade -r requirements.txt --no-user && npm run clean && npm run package"
     ],
     "extensionFile": "ms-toolsai-jupyter-insiders.vsix",
     "timeout": 30


extensions.json
```diff
@@ -797,10 +797,7 @@
   "ms-toolsai.jupyter": {
     "repository": "https://github.com/microsoft/vscode-jupyter",
     "custom": [
-      "npm ci",
-      "python3 -m pip --disable-pip-version-check install -t ./pythonFiles/lib/python --no-cache-dir --implementation py --no-deps --upgrade -r requirements.txt --no-user",
-      "npm run clean",
-      "npm run package"
+      "source ~/.nvm/nvm.sh && nvm use && nvm install-latest-npm && npm install --global @vscode/vsce && npm ci && python3 -m pip --disable-pip-version-check install -t ./pythonFiles/lib/python --no-cache-dir --implementation py --no-deps --upgrade -r requirements.txt --no-user && npm run clean && npm run package"
```
filiptronicek, line 800: 1. Can we separate the commands into lines, like it was before?
2. We can remove the nvm commands for sourcing and `nvm use`? Those should be automatically executed [[1](https://github.com/curbengh/openvsx-publish-extensions/blob/ae65295b43ba00a11c487e82c3f9a143615ef160/publish-extension.js#L68-L74)] - if that isn't working, please let me know, we should fix there. 
3. `vsce` will always be installed in the context of the publishing GitHub Action [[1]](https://github.dev/open-vsx/publish-extensions/blob/e5a086ac7ba52dfbcd80be3153cc27a9d5d44d12/.github/workflows/publish-extensions.yml#L45)

filiptronicek (commented):
Hey there @jared-nussbaum-frbb, thanks a lot for fixing this issue. 

Left a couple of comments below

jared-nussbaum-frbb (commented):
Made this by mistake. No need to rerun the Actions

extensions.json
jared-nussbaum-frbb: Thanks @filiptronicek. The issue does appear to be that although `nvm install` is run, it isn't used in the following exec commands for the custom block. You can see this in [this test run](https://github.com/jared-nussbaum-frbb/publish-extensions/actions/runs/4287026987/jobs/7467368863) that it still uses the system node version. The reasoning behind the format I had chosen is because the `nvm use` would only last for the one python exec shell session so I thought this would be better than sourcing and running `nvm use` each time.

jared-nussbaum-frbb: To summarize
1. We can have each major command like `npm ci`, `npm run package` on its own line. However we would need to resolve point 2 in order to avoid having to run `source ~/.nvm/nvm.sh && nvm use` at the start of those lines
2. We currently cannot get remove the nvm commands for sourcing and `nvm use`. While `nvm install` is run, there is never a switch over to the installed node version by nvm. There is also the issue where we can't just run `nvm use` once in the `custom` section because each command [[1](https://github.com/curbengh/openvsx-publish-extensions/blob/master/publish-extension.js#L72-L79)] is run in a separate bash session [[2](https://github.com/curbengh/openvsx-publish-extensions/blob/master/lib/exec.js#L24-L33)] which resets the node that runs to the system node version. Fixing this issue should remove the need for this PR.
3. vsce install can be taken out of this PR

Additionally to avoid the need to run `nvm install-latest-npm` it might make sense to run `nvm install` with the `--latest-npm` flag

Dealing with all these points will bring us back to the existing `custom` block so feel free to close/decline this PR in favor of another PR to address those points.

filiptronicek (approved):
Thanks a lot for explaining, we should get the fix in sooner rather than later, so I will merge. We can address some of my feedback later. 

Again, thanks a lot for taking care! 🙏

filiptronicek:
There we go: https://github.com/open-vsx/publish-extensions/actions/runs/4315544006/jobs/7530066249#step:9:535

PR #710, opened 2023-10-06
Update python to 3.8

apollo13:
There are multiple extensions currently failing since they apparently require python 3.8:

https://github.com/open-vsx/publish-extensions/actions/runs/6427077646/job/17452186756#step:11:5608

https://github.com/open-vsx/publish-extensions/actions/runs/6427077646/job/17452186756#step:11:5452

- [x] I have tried reaching out to the extension maintainers about publishing this extension to Open VSX (if not, please create an issue in the extension's repo using [this template](https://github.com/open-vsx/publish-extensions/blob/HEAD/docs/external_contribution_request.md)).

## Description

I have increased the minimal python version to 3.8, 3.7 is EOL and as such fails when building extensions due to missing dependencies etc…

/cc @filiptronicek

### DIFF
--- a/.github/workflows/publish-extensions.yml
+++ b/.github/workflows/publish-extensions.yml
@@ -39,7 +39,7 @@ jobs:
           bun-version: latest
       - uses: actions/setup-python@v4
         with:
-          python-version: "3.7"
+          python-version: "3.8"
       - uses: actions/setup-java@v3
         with:
           distribution: "microsoft"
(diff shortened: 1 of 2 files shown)


filiptronicek:
Thank you for reporting the extension issues and coming up with this PR. I have intended to raise it for quite a while now, because the black formatter extension was forcing us to stay on 3.7. 

I have committed a change to the validate workflow as well for the version bump, and we'll see how the publishing with 3.8 works: worst case we revert and see what's broken. 

Thank you again!

filiptronicek (approved):

apollo13:
Yeah not sure how we could handle that, isort currently also is on python 3.7 still (opened a PR there): https://github.com/microsoft/vscode-isort/pull/338

Will check black.

PR #736, opened 2024-01-08
Fix ms-python build

apollo13:
I extracted the required commands from https://github.com/microsoft/vscode-python/blob/main/.github/actions/build-vsix/action.yml

@filiptronicek Is there any good way to test this locally? Also when to use `prepublish` and when to use `custom`?

### DIFF
--- a/extensions.json
+++ b/extensions.json
@@ -854,7 +854,18 @@
   },
   "ms-python.python": {
     "repository": "https://github.com/microsoft/vscode-python",
-    "prepublish": "npx gulp installPythonLibs && python3 -m pip --disable-pip-version-check install packaging && python3 ./pythonFiles/install_debugpy.py && python3 ./pythonFiles/download_get_pip.py && python3 ./build/update_ext_version.py --release --for-publishing && npm run addExtensionPackDependencies && DISABLE_TRANSLATIONS=true npm run package",
+    "custom": [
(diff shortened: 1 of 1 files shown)


filiptronicek:
@apollo13 yes, although the repo is only tested on Ubuntu, it should also work on macOS. You can test a single extension as follows:
```
EXTENSIONS=ms-python.python SKIP_PUBLISH=true node publish-extensions
```

`SKIP_PUBLISH` makes sure it is only tested and not published.

apollo13:
Mhm it did build in CI (the failure is in Report results) and the size looks okay? I am having troubles building locally currently since I at least miss pyenv and others. Given that I copied 1:1 from the upstream action, it should work?

apollo13:
It also seems that 
```
python -m pip --disable-pip-version-check install packaging
```
runs in my default python3.12. Shouldn't pyenv create a 3.8 env before that and run it in there?

filiptronicek:
> Shouldn't pyenv create a 3.8 env before that and run it in there?

Thanks to the version property on the extension item, pyenv will download the version and set it as the default for the duration of the extension's build.

```json
"pythonVersion": "3.8",
```

image

filiptronicek (approved):
It indeed does build correctly, thank you! Merging this now

Issue #741, opened 2024-02-12
Add the `swyddfa.esbonio` extension


tchernobog:
Add the extension from https://marketplace.visualstudio.com/items?itemName=swyddfa.esbonio.

This is increasingly becoming a replacement for `lextudio.restructuredtext`.

filiptronicek:
Hey there @tchernobog, it seems the extension has been published in the meantime: https://open-vsx.org/extension/swyddfa/esbonio

PR #742, opened 2024-02-12
Add `swyddfa.esbonio` (#741)

tchernobog:
- [X] I have read the note above about PRs contributing or fixing extensions
- [X] I have tried reaching out to the extension maintainers about publishing this extension to Open VSX (if not, please create an issue in the extension's repo using [this template](https://github.com/open-vsx/publish-extensions/blob/HEAD/docs/external_contribution_request.md)).
- [X] This extension has an [OSI-approved OSS license](https://opensource.org/licenses) (we don't accept proprietary extensions in this repository)

## Description

Add `swyddfa.esbonio` (#741).

Author [has manifested interest](https://github.com/swyddfa/esbonio/issues/641) in publishing to OpenVSX once this hits version 1.0, but I would like a temporary solution for now as 1.0 might be several months away.

### DIFF
--- a/extensions.json
+++ b/extensions.json
@@ -813,7 +813,11 @@
       "vsce package --target=darwin-arm64",
       "vsce package --target=win32-x64"
     ],
-    "target": ["linux-x64", "darwin-arm64", "win32-x64"]
+    "target": [
+      "linux-x64",
+      "darwin-arm64",
+      "win32-x64"
+    ]
   },
   "ms-python.flake8": {
     "repository": "https://github.com/microsoft/vscode-flake8",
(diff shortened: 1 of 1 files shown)


filiptronicek (commented):
Hi @tchernobog 👋, thanks for the contribution.

I am happy to add the extension, but the size we build is about 180KB, while the official extension is somewhere about 1.4MB. I think there are some build steps we're missing

tchernobog:
Thanks @filiptronicek for catching this. I fixed the build, but I still have one problem.

As far as I can tell, the code in `open-vsx/publish-extensions` tries to infer the latest version of an extension using the github repository tags.

Unfortunately, the extension uses [a non-standard format](https://github.com/swyddfa/esbonio/tags). The latest tag is currently for an **older** release from a different branch than mainline (called here `develop`).

Here we need to consider only those tags that match `esbonio-vscode-extension-v*`.

During the build it's easy for me to do so, but as the context is already set at that point, the wrong version is published:

```json
{
  "context": {
    "msVersion": "0.11.0",
    "msLastUpdated": "2023-03-06T22:51:28.410Z",
    "msInstalls": 13902,
    "msPublisher": "swyddfa",
    "version": "0.11.0",
    "repo": "/tmp/repository/main",
    "ref": "esbonio-language-server-v0.16.4"
  }
}
```

How can this be fixed? I don't see anything in the JSON schema.

filiptronicek:
Thanks for looking into this further, @tchernobog. Does this mean that the currently built package for `esbonio-0.92.1.vsix` is incorrect, or did your git checkout tricks work out in the custom instructions?

The setup here looks quite complex and difficult to maintain, so I think it would be best to publish the current latest version manually and wait for the author to publish the 1.0 release themselves as you mentioned.

tchernobog:
@filiptronicek the results in term of package seem fine.

The only problem is that the deduced version seems wrong (0.11.0). If the result of this run is indeed `esbonio-0.92.1.vsix` then this is working well, if it deduces it as `esbonio-0.11.0.vsix` then it's wrong.

filiptronicek:
I see the issue now, thanks for pointing it out.

When publishing, this context is not used and the only version info, which is what is actually listed in the packaged extension (Open VSX takes care of that).

I am publishing v0.92.1 now

filiptronicek:
Published: https://open-vsx.org/extension/swyddfa/esbonio 🎉

PR #768, opened 2024-07-06
Manually Update SilverStripe Extension

### LINKED ISSUE #767: SilverStripe Extension is Incomplete on Open-VSX
I just noticed that the SilverStripe extension published on Open-VSX is missing the `/node_modules/coffeescript` directory in the `.vsix` file and thus draws the extension non-functioning. In this process, for whatever reason, other extensions break, too.

I have also noticed that the most recent version of the SilverStripe extension on the VSCode Marketplace is `v1.0.6` while the version hinted in the GitHub project is `v1.0.0`:

https://github.com/adrhumphreys/vscode-silverstripe/blob/a3fd311cdabdca6a23a3d5cc18c2bc00b7f16c2c/package.json#L10

Sadly, that project seems to be abandoned, so I don't think this is worth reporting to the original author.

manuth:
-   [ ] ~~I have tried reaching out to the extension maintainers about publishing this extension to Open VSX (if not, please create an issue in the extension's repo using [this template](https://github.com/open-vsx/publish-extensions/blob/HEAD/docs/external_contribution_request.md)).~~

## Description
As described in #767, the SilverStripe extension is unmaintained and hints the incorrect version number in its `package.json` file. Furthermore, the current `.vsix` file hosted on Open-VSX is broken and doesn't work properly - even causing other extensions to not getting loaded.

Therefore, changes made in this PR will:
  - Check out the (at time of writing) most recent commit
  - Set the (at time of writing) most recent version number in the `package.json` file
  - Apart from that publish it normally

Merging this PR will close #767

### DIFF
--- a/extensions.json
+++ b/extensions.json
@@ -15,7 +15,7 @@
   "adrianhumphreys.silverstripe": {
     "repository": "https://github.com/adrhumphreys/vscode-silverstripe",
     "custom": [
-      "git reset --hard a3fd311cdabdca6a23a3d5cc18c2bc00b7f16c2c",
+      "git checkout a3fd311cdabdca6a23a3d5cc18c2bc00b7f16c2c",
       "jq '.version = \"1.0.6\"' package.json > tmp.json && mv tmp.json package.json",
       "npm install",
       "NODE_OPTIONS=--openssl-legacy-provider vsce package -o extension.vsix"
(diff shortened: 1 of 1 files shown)


filiptronicek (approved):
Long time no see, @manuth! 

Thanks for this, I've tried the produced package locally, and it seems to work as expected. I'm going to merge this, and after it gets published remove the extension from the list, because it does not seem like it will get any further updates.

manuth:
Awesome to hear from you again! 😄 
tysm for taking care!

PR #803, opened 2024-12-31
chore: deprecate neocmake-lsp-vscode

Decodetalkers:
neocmakelsp is a single program, so the origin name is a little weired, so I have created another one instead, I want to deprecate the origin one and make user to use the neocmakelsp-vscode

-   [x] I have tried reaching out to the extension maintainers about publishing this extension to Open VSX (if not, please create an issue in the extension's repo using [this template](https://github.com/open-vsx/publish-extensions/blob/HEAD/docs/external_contribution_request.md)).

## Description

neocmakelsp is a single program, so the origin name is a little weired, so I have created another one instead, I want to deprecate the origin one and make user to use the neocmakelsp-vscode

### DIFF
--- a/extension-control/extensions.json
+++ b/extension-control/extensions.json
@@ -36,6 +36,13 @@
         }
     ],
     "deprecated": {
+        "decodetalkers.neocmake-lsp-vscode": {
+            "disallowInstall": false,
+            "extension": {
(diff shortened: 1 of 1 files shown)


filiptronicek (approved):
Hey there @Decodetalkers 👋!

PR looks great, let's get it shipped.

PR #823, opened 2025-02-06
Update extensions.json

ltrimino:
Add Ryan Osullivan extension for LS-DYNA FEA code. MIT license in the Repo

-   [ ] I have read the note above about PRs contributing or fixing extensions
-   [ ] I have tried reaching out to the extension maintainers about publishing this extension to Open VSX (if not, please create an issue in the extension's repo using [this template](https://github.com/open-vsx/publish-extensions/blob/HEAD/docs/external_contribution_request.md)).
-   [ ] This extension has an [OSI-approved OSS license](https://opensource.org/licenses) (we don't accept proprietary extensions in this repository)

## Description

This PR introduces X change for Y reason.

### DIFF
--- a/extensions.json
+++ b/extensions.json
@@ -1579,5 +1579,8 @@
   },
   "zxh404.vscode-proto3": {
     "repository": "https://github.com/zxh0/vscode-proto3"
+  },
+  "vscode_lsdyna": {
+    "repository": "https://github.com/osullivryan/vscode-lsdyna"
   }
 }


filiptronicek:
Hi @ltrimino 👋, looking at [the logs](https://github.com/EclipseFdn/publish-extensions/actions/runs/13187767759/job/36863245199?pr=823), it seems that the extension doesn't build (I suspect something broke in the past 4 years since it was last pushed to).

If you can get the extension to build (either with our instructions in `extensions.json` or with manual commands, I'd be happy to build it myself and one-time publish it (we generally don't add extensions to the list that don't receive updates, because it's easier to just manually publish them once).

ltrimino:
Hi @filiptronicek ,

thanks for giving it a go. I will see what I can do, fix that issue with the license and take a look at the manually building. BTW I do not see any set of particular instructions in the extensions. Json. Is there a wiki or some other document that I can look at?

filiptronicek:
@ltrimino the best place I can recommend for the structure of `extension.json` entries is either the [JSON schema](https://github.com/EclipseFdn/publish-extensions/blob/master/extensions-schema.json) or better yet, other entries in the `extensions.json` file, which offer examples for the properties defined in the schema.

Because we're going to be one-time publishing this, I'll be happy to run commands to get it packaged on my end and published. No need to adhere to `extensions.json`, it would be enough to share in a PR comment.

ltrimino:
@filiptronicek  I did took care of the license eca issue but I am at a lost as to how to build this thing. I was looking at both the json schema file and also the extensions.json file but it is not clear as to what needs to be done on my end. I did browse in the repository and there is a package.json ( https://github.com/osullivryan/vscode-lsdyna/blob/a2fa779b156e3d9cd31e12fa7a6ba326790adfe9/package.json ) file in the repository that maybe resembles what is needed for the build. Can you take a quick look and let me know if that is what you need?

filiptronicek:
@ltrimino thanks for signing the ECA. Essentially, what I'd like to ask of you are instructions on building the extension you're contributing. We can publish it, but need to build an extension package (a `.vsix` file) first.

Perhaps the best way to go forward with this would be if you could help me with compiling a list of commands to execute in the https://github.com/osullivryan/vscode-lsdyna/ repo to get an `extension.vsix`. If I'll be able to reproduce, I can publish the extension even without having to automate it via `extensions.json`.

PR #842, opened 2025-03-20
Update extensions.json

kinto0:
-   [x] I have tried reaching out to the extension maintainers about publishing this extension to Open VSX (if not, please create an issue in the extension's repo using [this template](https://github.com/open-vsx/publish-extensions/blob/HEAD/docs/external_contribution_request.md)).

## Description

Per suggestion [here](https://github.com/facebook/pyre-check/issues/975#issuecomment-2737561978), we would like our vscode extension available here.

### DIFF
--- a/extensions.json
+++ b/extensions.json
@@ -646,6 +646,9 @@
   "MehediDracula.php-namespace-resolver": {
     "repository": "https://github.com/MehediDracula/PHP-Namespace-Resolver"
   },
+  "meta.pyrefly": {
+    "repository": "https://github.com/facebook/pyrefly"
+  },
   "miguelsolorio.fluent-icons": {
     "repository": "https://github.com/misolori/vscode-fluent-icons"
   },


filiptronicek:
Hi @kinto0 👋, as per our [contributing guidelines](https://github.com/EclipseFdn/publish-extensions/blob/master/CONTRIBUTING.md#when-to-add-an-extension):
> A goal of Open VSX is to have extension maintainers publish their extensions [according to the documentation](https://github.com/eclipse/openvsx/wiki/Publishing-Extensions). The first step we recommend is to open an issue with the extension owner. If the extension owner is unresponsive for some time, this repo (publish-extensions) can be used as a temporary workaround to ensure the extension is published to Open VSX.

The above also includes owners, who should publish the extension from their respective repositories instead of this one, so that they own the publishing process from end to end and reduce the maintenance burden on this repository.

Let me know if you have any questions, happy to assist with setting up the pipeline on your end.

I also see that you already have a pipeline for publishing to the Microsoft Marketplace: https://github.com/facebook/pyre-check/blob/main/pyre2/.github/workflows/deploy_extension.yml. This is awesome, and means that setting up publishing to Open VSX (besides getting an access token as described in the [Wiki article](https://github.com/EclipseFdn/open-vsx.org/wiki/Publishing-Extensions)) should be as easy as using `ovsx` in the same fashion as `vsce` for publishing to https://open-vsx.org.

PR #845, opened 2025-03-24
Update extensions.json - deprecate Azure Repos extension

tfroment:
## Description

This PR introduces deprecation of https://open-vsx.org/extension/ms-vsts/team ("Azure Repos extension") extension to fix https://github.com/EclipseFdn/open-vsx.org/issues/3635

### DIFF
--- a/extension-control/extensions.json
+++ b/extension-control/extensions.json
@@ -434,6 +434,13 @@
                 "id": "KylinIdeTeam.kylin-clangd",
                 "displayName": "Kylin Clangd"
             }
+        },
+        "ms-vsts.team": {
+            "disallowInstall": false,
+            "extension": {
+                "id": "ms-vsts.team",
+                "displayName": "Azure Repos"
+            }
         }
     },
     "migrateToPreRelease": {


tfroment (commented):
Hi @filiptronicek:  Any idea why Sonar execution fails here? https://github.com/EclipseFdn/publish-extensions/actions/runs/14039850728/job/39306996768 it looks like it fails with all the jobs: https://github.com/EclipseFdn/publish-extensions/actions/workflows/sonar.yml

filiptronicek (approved):
Deprecation LGTM, thanks!

filiptronicek:
@tfroment Sonar seems to be complaining about a missing token. We are seemingly [referencing it](https://github.com/EclipseFdn/publish-extensions/actions/runs/14039850728/workflow#L53) in the workflows, but it's not clear to me if it's actually set up in the repository secrets. 

Do you by chance have admin access to the repository and hence its settings? I'm asking because I personally don't and cannot check if it's really missing.
If I'll find some time in the coming weeks, I'll look into replacing the action we're using now  (`sonarsource/sonarcloud-github-action`) with `sonarqube-scan-action`, as the deprecation warning recommends.

PR #881, opened 2025-05-19
Use separate jobs instead of child_process

amvanbaren:
This PR introduces the use of separate jobs instead of child_process to sandbox the extension build process.

### DIFF
--- a/local-workflow.js
+++ b/local-workflow.js
@@ -13,7 +13,12 @@ const publishExtensionsScript = require("./scripts/publish-extensions");
 const buildExtensionScript = require("./scripts/build-extension");
 const publishExtensionScript = require("./scripts/publish-extension");
 
-await publishExtensionsScript(async (extension, publishContext) => {
-    const extensionFiles = await buildExtensionScript(extension, publishContext);
-    await publishExtensionScript(extension.id, extensionFiles);
-})
+(async () => {
+    process.env.EXTENSIONS = "dfinity-foundation.vscode-motoko";
+    process.env.SKIP_PUBLISH = "false";
+    process.env.FORCE = "true";
+    await publishExtensionsScript(async (extension, publishContext) => {
+        const extensionFiles = await buildExtensionScript(extension, publishContext);
+        await publishExtensionScript(extension.id, extensionFiles);
+    });
+})();
--- a/local-workflow.js
+++ b/local-workflow.js
@@ -14,9 +14,9 @@ const buildExtensionScript = require("./scripts/build-extension");
 const publishExtensionScript = require("./scripts/publish-extension");
 
 (async () => {
-    process.env.EXTENSIONS = "dfinity-foundation.vscode-motoko";
-    process.env.SKIP_PUBLISH = "false";
-    process.env.FORCE = "true";
+    process.env.SKIP_PUBLISH ??= "true";
+    process.env.FORCE ??= "true";
+
     await publishExtensionsScript(async (extension, publishContext) => {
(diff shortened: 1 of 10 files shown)


tfroment (approved):
+1

tfroment:
hi @amvanbaren, do you know why this is failed https://github.com/EclipseFdn/publish-extensions/actions/runs/15301776884/job/43043958569?pr=881

amvanbaren:
The reporting functionality is broken: https://github.com/EclipseFdn/publish-extensions/blob/7420c176b98dc167ffd3020a961d6b0606367c69/.github/workflows/publish-extensions.yml#L50-L77

There's a check for important extensions: https://github.com/EclipseFdn/publish-extensions/blob/f7cfb5c517bbf53d772704383db8b22b4cb895aa/report-extensions.ts#L226C1-L232C2

Because the reporting functionality is broken every extension is seen as outdated and the Validate PR job fails.

scripts/build-extension.js
filiptronicek, line 1: Nice, I like finally not having everything in the root 😄

filiptronicek, line 1: Can you please run `npm run format` here?

filiptronicek (commented):
In general, I'm aligned and happy with this approach --> we never really did publishing outside of GitHub Actions anyway.

I'm curious about:
- How does our local testing story look like? I used to run `node publish-extension.js` with a list of extensions when debugging build issues with certain extensions
- How can we test the PR before landing it? If it's not worth the effort to do here, I'm also fine with merging it and testing then
- Do we have an idea on how this will affect timing? I know there's some concurrency limits for GHA workflows, so given there's > 400 extensions, I worry the publishing times might get lengthy.

scripts/download-extension.js
```diff
@@ -0,0 +1,38 @@
...
+const resolveExtension = require("../lib/resolveExtension").resolveExtension;
+
+module.exports = async () => {
+    const extension = JSON.parse(process.env.EXTENSION);
+    console.debug(`Resolving downloads for ${extension.id}`);  
+
+    const publishContext = JSON.parse(process.env.PUBLISH_CONTEXT);
+    publishContext.msLastUpdated = new Date(publishContext.msLastUpdated);
+    publishContext.ovsxLastUpdated = new Date(publishContext.ovsxLastUpdated);
+    await resolveExtension(
```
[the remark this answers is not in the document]
amvanbaren: Yes, it downloads extension files to `/tmp/download/`

amvanbaren:
> Do we have an idea on how this will affect timing? I know there's some concurrency limits for GHA workflows, so given there's > 400 extensions, I worry the publishing times might get lengthy.

Yes, it's going to be lengthy and create a lot of jobs: from https://github.com/amvanbaren/publish-extensions/actions?page=2 to https://github.com/amvanbaren/publish-extensions/actions?page=18 for a single run.

local-workflow.js
```diff
@@ -0,0 +1,24 @@
...
+
+// @ts-check
+const publishExtensionsScript = require("./scripts/publish-extensions");
+const buildExtensionScript = require("./scripts/build-extension");
+const publishExtensionScript = require("./scripts/publish-extension");
+
+(async () => {
+    process.env.EXTENSIONS = "dfinity-foundation.vscode-motoko";
+    process.env.SKIP_PUBLISH = "false";
+    process.env.FORCE = "true";
```
filiptronicek, line 19: I'd prefer to only set these if they're not set by the executing context. Also, let's default to not settings extensions and since it's the local workflow for debugging, skip publishing by default.

PR #885, opened 2025-05-26
Address swift extension rename

filiptronicek:
Publishes the extension under the new namespace.

Also deprecated the old in favor of the renamed version.

Thanks @NSHaoSong for pointing this out in https://github.com/EclipseFdn/publish-extensions/pull/721#issuecomment-2910380807!


NSHaoSong:
Looks like we can close this now. 

We have the official extension published already: https://open-vsx.org/extension/swiftlang/swift-vscode

PR #912, opened 2025-07-01
[optify] Add Optify extension

juharris:
-   [ ] I have tried reaching out to the extension maintainers about publishing this extension to Open VSX (if not, please create an issue in the extension's repo using [this template](https://github.com/open-vsx/publish-extensions/blob/HEAD/docs/external_contribution_request.md)). _I am the maintainer, but I want the extension to be automatically published because this feels more legitimate._

## Description
Add Optify extension.

I followed the guidelines in https://github.com/EclipseFdn/open-vsx.org/wiki/Auto-Publishing-Extensions#request-an-extension-be-auto-published

I have [a GitHub action](https://github.com/juharris/optify/blob/main/.github/workflows/vsc_extension-build_test.yml) to publish it to the VS Marketplace. Here is the published extension: https://marketplace.visualstudio.com/items?itemName=optify-config.optify

I didn't realize that extensions could be automatically published to Open VSX so I already manually published this one: https://open-vsx.org/extension/optify-config/optify
but I would rather it get automatically published and look more legit, I hope that's alright.

### DIFF
--- a/extensions.json
+++ b/extensions.json
@@ -1162,6 +1162,10 @@
   "openfl.lime-vscode-extension": {
     "repository": "https://github.com/openfl/lime-vscode-extension"
   },
+  "optify-config.optify": {
+    "repository": "https://github.com/juharris/optify",
+    "location": "vscode/extension"
+  },
   "paiqo.databricks-vscode": {
     "repository": "https://github.com/paiqo/Databricks-VSCode"
   },


juharris:
Thanks for getting back to me so quickly @filiptronicek !
I think it's okay now because https://open-vsx.org/extension/optify-config/optify has the checkmark and doesn't have the scary warning at the top now. I guess that's because I managed to claim the namespace, which was pretty easy as the issue was automatically resolved.
I was looking at other extensions, like this Ruby pack: https://open-vsx.org/extension/Shopify/ruby-extensions-pack which looks pretty sketchy whereas the Python one looks super legit cause it says it's published by open-vsx: https://open-vsx.org/extension/ms-python/python
So I kind of wanted mine to look more like the Python one.

juharris:
One reason I left Microsoft to go to Shopify was to be at a smaller faster paced company hehe
I'll ask around to see what's going on with that Ruby pack especially because the owner, Andy, doesn't work at Shopify anymore! Anyway, I don't think this is a big company thing, Andy probably just published it in the easiest way possible and didn't bother setting up a namespace.

PR #810, opened 2025-01-15
feat(extensions.json): Boundary.baml-extension-preview should redirect to Boundary.baml-extension

sxlijin:
Per [docs](https://github.com/EclipseFdn/open-vsx.org/wiki/Managing-Extensions#deprecating-an-extension), I'm requesting a deprecation of https://open-vsx.org/extension/Boundary/baml-extension-preview.

[Boundary.baml-extension-preview](https://open-vsx.org/extension/Boundary/baml-extension-preview) was published while doing some internal alpha testing of our open vsx release process, but because I chose such a similar name to our actual extension, [Boundary.baml-extension](https://open-vsx.org/extension/Boundary/baml-extension), it looks like some of our users have landed on baml-extension-preview instead of baml-extension.

Deprecation seems like the correct way to handle this; please let me know if there's a more appropriate way to do so.

PR #977, opened 2025-09-04
feat(extensions.json): temporarily restore Boundary.baml-extension-preview

sxlijin:
Boundary.baml-extension-preview still comes up in Cursor search results (which are backed by open vsx afaik) presumably because (1) the current name is "Baml (preview2)" and (2) some of the associated tags/categories say "baml"

I'd like to temporarily restore baml-extension-preview so that we can publish a new version free of any metadata, essentially "nuking" it for any users coming to the extension, and then re-deprecate it. Let me know if this sounds reasonable (I recognize that Eclipse doesn't have any control over how Cursor interfaces with the openvsx registry).

cc @filiptronicek as reviewer on #810

### DIFF
--- a/extension-control/extensions.json
+++ b/extension-control/extensions.json
@@ -443,13 +443,6 @@
                 "displayName": "SerenityOS DSL Syntax Highlight"
             }
         },
-        "Boundary.baml-extension-preview": {
-            "disallowInstall": true,
-            "extension": {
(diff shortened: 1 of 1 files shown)


sxlijin:
Hi, yes, this is still relevant.

Please let me know if I should do something to re-trigger CI, I can't see the failure logs anymore.

PR #988, opened 2025-09-08
Update extensions.json

Anurag-Kochar-1:
This PR deprecates my old extension [that-typescript-extension](https://open-vsx.org/extension/AnuragKochar/that-typescript-extension).

I have renamed the extension, updated the icon, and published the new version under [better-not-any](https://open-vsx.org/extension/AnuragKochar/better-not-any), which should be used going forward.

The old extension should be marked as deprecated and point to the new one as its replacement.

### DIFF
--- a/extension-control/extensions.json
+++ b/extension-control/extensions.json
@@ -62,6 +62,13 @@
         }
     ],
     "deprecated": {
+        "AnuragKochar.that-typescript-extension" : {
+            "disallowInstall": true,
+            "extension": {
+                "id": "AnuragKochar.that-typescript-extension",
+                "displayName": "that-typescript-extension"
+            }
+        },
          "HerrInformatiker.ai-rules-sync" : {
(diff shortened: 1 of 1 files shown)


extension-control/extensions.json
```diff
@@ -62,6 +62,13 @@
         }
     ],
     "deprecated": {
+        "AnuragKochar.that-typescript-extension" : {
+            "disallowInstall": true,
+            "extension": {
+                "id": "AnuragKochar.that-typescript-extension",
```
filiptronicek, line 68: Hi there @Anurag-Kochar-1 👋, thanks for your PR. The field under `extension.id` should be used as the replacement (if applicable), so please either remove it or change to your preferred replacement.

I also notice the extension is not on https://open-vsx.org anymore, so please note that this will only affect existing users who should see an in-editor nudge towards migrating or uninstalling.

PR #1006, opened 2025-09-29
Deprecate modular-mojotools.vscode-mojo-nightly

AmaranthineCodices:
-   [ ] I have tried reaching out to the extension maintainers about publishing this extension to Open VSX (if not, please create an issue in the extension's repo using [this template](https://github.com/open-vsx/publish-extensions/blob/HEAD/docs/external_contribution_request.md)).

## Description

We're discontinuing the nightly extension in favor of the stable one.


zbowling:
I left modular, but yeah, modular moved a single extension instead of a stable and nightly. I had originally published both before I left the company. Removing the nightly makes sense!

PR #1053, opened 2025-12-04
deprecated Greenplumwine.mybatis-helper

Greenplumwine:
deprecated Greenplumwine.mybatis-helper

### DIFF
--- a/.github/workflows/publish-extension.yml
+++ b/.github/workflows/publish-extension.yml
@@ -1,4 +1,5 @@
 name: Publish extension
+run-name: Publish extension ${{ fromJson(inputs.extension).id }}
 on:
   workflow_dispatch:
     inputs:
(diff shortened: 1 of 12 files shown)


Greenplumwine:
> 嗨 👋 感谢你的贡献！我可以请你签署贡献者许可协议吗？我们需要它才能合并这个。你应该能在下面的 PR 检查中找到一个链接

我应该已经完成了 ECA

netomi:
may I suggest that you just remove the extension as a whole? I see that you published the same extension with a new extension id, but the extension has ~300 downloads, adding a deprecation notice with a redirect does not really make much sense imho.

Greenplumwine:
> may I suggest that you just remove the extension as a whole? I see that you published the same extension with a new extension id, but the extension has ~300 downloads, adding a deprecation notice with a redirect does not really make much sense imho.

Okay, I agree to delete this plugin

filiptronicek:
Hi there @Greenplumwine 👋. Haven't had a chance to look at whether your extension is published or not, but if the intent is to deprecate it, I am happy to go ahead with the process.
Deprecation often makes sense even after removal, since editors will suggest users who already have the extension installed to migrate over to the new one.

If you'd like to proceed, please rebase with the latest upstream `master` branch

PR #1074, opened 2025-12-27
deprecate phpcs in favor of vscode-phpcs

JohnRDOrazio:
## Description

This PR deprecates all previously published `phpcs` extensions in favor of the newly published extension `vscode-phpcs`, seeing that the previously published extensions are currently unmaintained. The name change from `phpcs` to `vscode-phpcs` was required for the VSCode Marketplace, where it was not possible to publish under the same name (perhaps because the previously published extensions were not marked as deprecated).

### DIFF
--- a/extension-control/extensions.json
+++ b/extension-control/extensions.json
@@ -736,7 +736,8 @@
             "extension": {
                 "id": "johnrdorazio.vscode-phpcs",
                 "displayName": "vscode-phpcs"
-            }
+            },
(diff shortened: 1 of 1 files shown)


JohnRDOrazio:
@filiptronicek  I have removed the references to `ikappas.phpcs` and `shevaua.phpcs`, the only remaining reference is to my own `johnrdorazio.phpcs` deprecated in favour of `johnrdorazio.vscode-phpcs`.

PR #1077, opened 2026-01-08
feat: add repository for invisible-squiggles

michen00:
## Description

This PR introduces the [invisible-squiggles](https://marketplace.visualstudio.com/items?itemName=michen00.invisible-squiggles) extension for broader compatibility with popular IDEs.

PR #1078, opened 2026-01-08
feat: add repository for invisible-squiggles

michen00:
## Description

This PR introduces the [invisible-squiggles](https://marketplace.visualstudio.com/items?itemName=michen00.invisible-squiggles) extension for broader compatibility with popular IDEs.

Related: #1077

### DIFF
--- a/extensions.json
+++ b/extensions.json
@@ -648,6 +648,9 @@
   "MehediDracula.php-namespace-resolver": {
     "repository": "https://github.com/MehediDracula/PHP-Namespace-Resolver"
   },
+  "michen00.invisible-squiggles": {
+    "repository": "https://github.com/michen00/invisible-squiggles"
+  },
   "miguelsolorio.fluent-icons": {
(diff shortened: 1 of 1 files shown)


michen00:
@kineticsquid @netomi , is there anything else I should do to get this merged? thanks very much 🙏

michen00:
thank you, I will take a look

michen00:
Withdrawing this — happily obsolete. As of v0.4.0 the extension is self-published to Open VSX: https://open-vsx.org/extension/michen00/invisible-squiggles

The repository now publishes to Open VSX and the VS Code Marketplace from one reproducible, provenance-attested CI build, so the auto-publish entry this PR asked for would duplicate a publisher that already exists. That is the outcome CONTRIBUTING.md points at — "a goal of Open VSX is to have extension maintainers publish their extensions" — so the right move is to take this off your queue rather than leave it pending.

Thanks for maintaining the fallback route. It is a good thing to have existed while I had not gotten around to publishing properly.

netomi:
perfect

PR #1169, opened 2026-07-20
Deprecate CicerBro.phpantom → phpantom.phpantom

CicerBro:
## Summary
- Mark `CicerBro.phpantom` as deprecated on Open VSX
- Set `disallowInstall: true` so new installs are blocked
- Recommend replacement `phpantom.phpantom` (official PHPantom extension)

I am the publisher of `CicerBro.phpantom`. Version 0.3.3 is already published with an in-editor migration prompt; this control-file entry makes the deprecation official in the Open VSX UI (applied by the nightly batch job after merge).

### DIFF
--- a/extension-control/extensions.json
+++ b/extension-control/extensions.json
@@ -950,6 +950,14 @@
             },
             "additionalInfo": "MathWorks now publishes and maintains the official MATLAB extension on Open VSX. Please migrate to mathworks.language-matlab."
         },
+        "CicerBro.phpantom": {
+            "disallowInstall": true,
+            "extension": {
+                "id": "phpantom.phpantom",
+                "displayName": "PHPantom"
+            },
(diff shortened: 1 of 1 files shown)


netomi:
can you sign the ECA? Details available at https://api.eclipse.org/git/eca/status/gh/EclipseFdn/publish-extensions/1169

CicerBro:
You want me to sign something **and** attach my legal name to this account **and** also my address? Just to mark my own extension as deprecated now an official one exists?

You must be on crack.

netomi:
Thanks for your kind words, its always appreciated when receiving positive feedback from the community for this free service.

I was wrongly assuming that as you already have an Eclipse account which is a requirement to publish an extension to open-vsx.org, clicking the button to sign an ECA would be straight-forward, but apparently I was wrong in my judgement.

